PR #10, opened 2019-11-08
SRS extension update

mattdurak:
This adds support to the SRS extension to find prefixes that have numbers in them
Also modifies the extension to the newer template and adds a few unit tests for the regular expressions


srs_extension/test/suite/getprefixes.unit.test.js
```diff
@@ -0,0 +1,79 @@
...
+    test('Get Prefixes with one prefix multiple uses', () => {
+        var prefixes = reqparser.getPrefixes(example2);
+        assert.equal(prefixes.length, 1);
+        assert.equal(prefixes[0], 'SRS_CODE_REQ_');
+    });
+
+    test('Get Prefixes with numbers', () => {
+        var prefixes = reqparser.getPrefixes(exampleWithNumbers);
+        assert.equal(prefixes.length, 1);
+        assert.equal(prefixes[0], 'SRS_2_CODE_2_FURIOUS_');
```
dcristoloveanu: >_FURIOUS_' [](start = 47, length = 10)

:-)

srs_extension/test/runTest.js
```diff
@@ -0,0 +1,23 @@
...
+        // Passed to --extensionTestsPath
+        const extensionTestsPath = path.resolve(__dirname, './suite/index');
+
+        // Download VS Code, unzip it and run the integration test
+        await runTests({ extensionDevelopmentPath, extensionTestsPath });
+    } catch (err) {
+        console.error('Failed to run tests');
+        process.exit(1);
+    }
+}
```
dcristoloveanu: We could run this as part of the gate. Consider.

mattdurak: Giving this a try (it seems to run from a command prompt on my machine so we'll see if it works in the gate too)

PR #18, opened 2020-04-08
Add support to srs extension for tagging multiple lines

mattdurak:
Adds command to SRS extension with alt-F9 to tag all lines that are selected (preserving mark-down lists, but otherwise just tags everything as-is). Version 0.0.8

### DIFF
--- a/srs_extension/README.md
+++ b/srs_extension/README.md
@@ -34,7 +34,7 @@ To insert a new requirement tag:
 
 All requirement tags use the following syntax: `SRS_SOMESTRING_<DEVID>_<REQID>`
 
-You may also insert requirement tags for the currently selected text b y pressing `Alt+F9`. This will:
+You may also insert requirement tags for the currently selected text by pressing `Alt+F9`. This will:
  - Tag all non-empty (white space only) lines
  - Preserve unordered list markdown (i.e. skip over combinations of `-` and space)
 
--- a/srs_extension/README.md
+++ b/srs_extension/README.md
@@ -34,6 +34,10 @@ To insert a new requirement tag:
 
 All requirement tags use the following syntax: `SRS_SOMESTRING_<DEVID>_<REQID>`
 
+You may also insert requirement tags for the currently selected text b y pressing `Alt+F9`. This will:
+ - Tag all non-empty (white space only) lines
+ - Preserve unordered list markdown (i.e. skip over combinations of `-` and space)
+
 (Right now the requirement tag format is hardcoded.)
 
 There are a couple more things you can do, by bringing up the command pane (by pressing F1):
--- a/srs_extension/extension.js
+++ b/srs_extension/extension.js
@@ -18,26 +18,32 @@ function activate(context) {
         if(id && id.trim().length > 0) {
             devId = id;
             saveDevId(id);
+            return true;
         }
+        return false;
     };
 
     var setReqPrefix = function (prefix) {
         if(!!prefix) {
             if(prefix === ADD_NEW_PREFIX) {
-                vscode.window.showInputBox({
+                return vscode.window.showInputBox({
                     placeHolder: 'New requirement prefix',
                     prompt: 'Enter the new requirement prefix to use'
                 })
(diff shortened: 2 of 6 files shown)


srs_extension/README.md
```diff
@@ -34,6 +34,10 @@ To insert a new requirement tag:
 
 All requirement tags use the following syntax: `SRS_SOMESTRING_<DEVID>_<REQID>`
 
+You may also insert requirement tags for the currently selected text b y pressing `Alt+F9`. This will:
```
dcristoloveanu, line 37: >b y [](start = 69, length = 3)

by

srs_extension/extension.js
```diff
@@ -105,6 +105,80 @@ function activate(context) {
...
+        var text = loadText();
+
+        if (text) {
+            var nextReqTag = reqParser.getNextReq(text, reqPrefix, devId);
+
+            var editor = vscode.window.activeTextEditor;
+            var document = editor.document;
+
+            var endSnippet = ' **]**';
+
```
dcristoloveanu, line 118: Do you want to add any tests for these? Seems like maybe the part after loadtext can actually be tested easily.

srs_extension/extension.js
```diff
@@ -105,6 +105,80 @@ function activate(context) {
         }
     };
 
+    var insertReqTags = function () {
+        var text = loadText();
+
+        if (text) {
```
dcristoloveanu, line 111: >if (text) { [](start = 8, length = 11)

Maybe there should be a method that gets as arg the input text, returns the text as it should be pasted in the editor.
Then it would be testable.

dcristoloveanu: At least we should add an issue with this so we can track it somehow ...

---
In reply to: [405873254](https://github.com/Azure/azure-c-build-tools/pull/18#discussion_r405873254) [](ancestors = 405873254)

dcristoloveanu: Or even better maybe you can refactor the code a bit now to have that method even if you do not add all the tests for it now ... Althoughonce refactored my feeling is that adding some tests might be very easy :-).

---
In reply to: [405873414](https://github.com/Azure/azure-c-build-tools/pull/18#discussion_r405873414) [](ancestors = 405873414,405873254)

srs_extension/extension.js
mattdurak: I figured out how to get int tests working so I added those and I have a few tests for this and the single requirement command :)

---
In reply to: [405872908](https://github.com/Azure/azure-c-build-tools/pull/18#discussion_r405872908) [](ancestors = 405872908)

srs_extension/extension.js
mattdurak: Int tests added (did not refactor this for now)

---
In reply to: [405873662](https://github.com/Azure/azure-c-build-tools/pull/18#discussion_r405873662) [](ancestors = 405873662,405873414,405873254)

srs_extension/test/suite/index.js
```diff
@@ -5,7 +5,8 @@ const glob = require('glob');
 function run() {
     // Create the mocha test
     const mocha = new Mocha({
-        ui: 'tdd'
+        ui: 'tdd',
+        timeout: 600000
```
dcristoloveanu, line 9: >600000 [](start = 17, length = 6)

600s? Oh well, does not matter anyway :-).

srs_extension/test/suite/editor.int.test.js
```diff
@@ -0,0 +1,347 @@
...
+ - **SRS_MY_PREFIX_08_002: [** do something **]**
+
+ - **SRS_MY_PREFIX_08_003: [** do something else **]**
+
+   - **SRS_MY_PREFIX_08_004: [** sublist item **]**
+
+**SRS_MY_PREFIX_08_005: [** Last one **]**
+`;
+
+suite('Editor Int Test Suite', () => {
```
dcristoloveanu, line 154: >Editor Int Test Suite', () => { [](start = 7, length = 31)

I love the fact we have some tests now.

PR #32, opened 2020-07-21
have a BINARY_DIR option

anporumb:


build_functions/CMakeLists.txt
```diff
@@ -207,12 +207,31 @@ function(build_test_folder test_folder)
...
+        set(PARSING_BINARY_DIR OFF) #see https://cmake.org/cmake/help/latest/command/add_subdirectory.html for "binary_dir"
+        
+        foreach(f ${ARGN}) 
+            set(skip_to_next FALSE)
+            if(${f} STREQUAL "BINARY_DIR")
+                SET(PARSING_BINARY_DIR ON)
+                set(skip_to_next TRUE)
+            endif()
+
+            if(NOT skip_to_next)
```
mattdurak, line 221: >skip_to_next [](start = 19, length = 12)

I read this a few times and got confused, couldn't this just be an "else" for the condition above?

```
if (string "BINARY_DIR")
    PARSING_BINARY_DIR = true
else if (PARSING_BINARY_DIR)
    BINARY_DIR=...
endif
```

anporumb: I agree that the proposed code looks better and it is easier to read. Changed.

---
In reply to: [458434515](https://github.com/Azure/azure-c-build-tools/pull/32#discussion_r458434515) [](ancestors = 458434515)

PR #36, opened 2020-09-03
Add wd4702 for MSC for release configs in order to be able to build with LTCG

dcristoloveanu:
Add wd4702 for MSC for release configs in order to be able to build with LTCG


mattdurak (approved):
:shipit:

PR #46, opened 2020-11-03
Add debugging info for Linux builds

dcristoloveanu:
Add debugging info for Linux builds

### DIFF
--- a/build_functions/CMakeLists.txt
+++ b/build_functions/CMakeLists.txt
@@ -604,8 +604,8 @@ macro(set_default_build_options)
         string(REGEX REPLACE "/W[1-3]" "/W4" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
         
     elseif(UNIX) #LINUX OR APPLE
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -g")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror -g")
     endif()
 
     if(${run_valgrind})


build_functions/CMakeLists.txt
```diff
@@ -604,8 +604,8 @@ macro(set_default_build_options)
         string(REGEX REPLACE "/W[1-3]" "/W4" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
         
     elseif(UNIX) #LINUX OR APPLE
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -g")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror -g")
```
mattdurak, line 608: We could also change the build type in the script to be "-DCMAKE_BUILD_TYPE=Debug" instead of defaulting to Release (since the Makefile generator bakes in the type unlike the VS one). But I guess this means we will always use essentially RelWithDebInfo on Linux so maybe that's acceptable.

mattdurak (approved):
:shipit:

build_functions/CMakeLists.txt
jebrando: If you add the -g here, there will be no way to override it, shouldn't this be optional?  Maybe default it, but make it so we can remove it or add to it.

jebrando (approved):

anporumb (approved):
:shipit:

build_functions/CMakeLists.txt
dcristoloveanu: We could. But maybe "when time comes".

---
In reply to: [516905216](https://github.com/Azure/c-build-tools/pull/46#discussion_r516905216) [](ancestors = 516905216)

dcristoloveanu: Also "when time comes".

---
In reply to: [516934619](https://github.com/Azure/c-build-tools/pull/46#discussion_r516934619) [](ancestors = 516934619)

PR #48, opened 2020-11-04
Split valgrind options

dcristoloveanu:
Split valgrind options so that we can select whether to run DRD or not (DRD does not like at all futexes)


mattdurak (approved):
:shipit:

PR #50, opened 2020-12-04
Output folder update

dcristoloveanu:
Each test project should place its binaries in a subfolder of the designated cmake output folder.


mattdurak (approved):
:shipit:

PR #52, opened 2021-01-11
Do not build traceability tool when run_traceability is OFF

dcristoloveanu:
Do not build traceability tool when run_traceability is OFF


mattdurak (approved):
:shipit:

PR #53, opened 2021-01-19
Add defaults for CMAKE_RUNTIME_OUTPUT_DIRECTORY_*

mattdurak:
Can be overridden before or after set_default_build_options but by default, binaries go in the normal cmake binary directory


jebrando (approved):
:shipit:

dcristoloveanu (approved):
:shipit:

PR #54, opened 2021-01-20
Need to always create traceabilitytool target

mattdurak:
Otherwise when we include this cmakelist with run_traceabilitytool off we don't get the target and builds fail


anporumb (approved):

CMakeLists.txt
```diff
@@ -31,7 +31,5 @@ option(run_traceability "run traceability tool (default is ON)" ON)
 add_subdirectory(build_functions)
 add_subdirectory(test_projects)
 if(WIN32)
-    if(${run_traceability})
```
dcristoloveanu, line 34: >${run_traceability}) [](start = 7, length = 20)

erm, I think this is a bit of a problem.
I am building with Ninja for the iwyu. And that does not have C#.
So I put on purpose this guard ...
So when building with Ninja we should not have this target => the consumers of c-build-tools will have to look at run_traceability and not add the traceability command

dcristoloveanu (changes requested):
:clock1:

CMakeLists.txt
mattdurak: what about guarding on the visual studio generator here? If that's possible.... Otherwise all other repos will have to make sure run_traceability stays with the given value for c-build-tools, but turn it off for the other deps.

---
In reply to: [561141751](https://github.com/Azure/c-build-tools/pull/54#discussion_r561141751) [](ancestors = 561141751)

mattdurak: Would if (${MSVC}) work and get skipped for Ninja?

---
In reply to: [561144091](https://github.com/Azure/c-build-tools/pull/54#discussion_r561144091) [](ancestors = 561144091,561141751)

dcristoloveanu (approved):
:shipit:

PR #55, opened 2021-01-20
Add basic rules for iwyu

dcristoloveanu:
- Add basic rules for iwyu
- Guard manifest embedding by generator being Visual Studio


mattdurak (approved):
:shipit:

PR #80, opened 2021-10-01
Add tool for checking real symbols

parth21999:
Add tool to check if given lib contains `[symbol]` and also `real_[symbol]`.


real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+function Check-Symbols {
```
mattdurak, line 3: consider adding [comment based help](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-7.1#examples)

mattdurak:
My only other comment is that we should run your test in the gate. Maybe a powershell script (or whatever you think) that runs your tool against the 3 different libs you build in the test directory and checks that the output is expected.

New tests like that would need to be defined as devops tasks (see [https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops) and we have lots of examples of doing this in block_store's build directory)
And would be defined in the build folder with the other devops task yaml.

Probably fine as a separate task/PR if you want, but I do think we should validate your test in the gate.

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+    )   
+        try {
+            [string]$output = cmd.exe /c dumpbin /ALL $lib
+        }
+        catch {
+            Write-Error  "Unable to run dumpbin on given lib"
+            return 2
+        }
+        $start_token = "public symbols"
+        $end_token = "Archive member name"
```
mattdurak, line 15: Are these tokens documented publicly? Wondering if the binary layout could change at some point and we wouldn't know (or would want some kind of version check).

mattdurak (approved):
:shipit:

real_check/real_check.ps1
parth21999: I could not find any public documentation on dumpbin's output format. Might be a good idea to fix dumpbin version.

real_check/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+add_subdirectory(test)
```
dcristoloveanu, line 3: Nitpick: consider adding a new line before add_subdirectory for readability.

real_check/real_check.md
```diff
@@ -0,0 +1,10 @@
+# real_check 
+
+This tool checks if the given library contains a pair of symbols of the form `[symbol]` and `real_[symbol]`. If the given library contains no such pair, the tool returns exit code `0`. If the given library contains such a pair, the tool prints these symbols to stderror and returns exit code `1`. If the tool is unable to determine if the library contains such a pair, it returns exit code `2`.
```
dcristoloveanu, line 3: >the [](http://example.com/codeflow?start=20&length=3)

a

dcristoloveanu, line 3: >urns exit co [](http://example.com/codeflow?start=277&length=12)

Can we split the text over a couple of lines. Usually one sentence per line makes it easy enough to look at.

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+function Check-Symbols {
+    param(
+        [string] $lib
+    )   
+        try {
+            [string]$output = cmd.exe /c dumpbin /ALL $lib
+        }
+        catch {
+            Write-Error  "Unable to run dumpbin on given lib"
```
dcristoloveanu, line 11: >given lib [](http://example.com/codeflow?start=51&length=9)

We should print the lib name

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+            [int] $end = $output.IndexOf($end_token, $start)
+            $symbol_text = ($output.Substring($start + $start_token.Length+1, $end-$start-$start_token.Length-1))
+            $split_symbols = $symbol_text.Split('',[StringSplitOptions]::RemoveEmptyEntries)
+        }
+        catch {
+            Write-Error  "Unable to parse dumpbin output"
+            return 2
+        }
+        if($split_symbols.Length -lt 2) {
+            Write-Error  "Unable to parse dumpbin output"
```
dcristoloveanu, line 27: I would love to have some comments with an example of what this does.
Like:
"dumpbin gives us this format: 2 columns containing symbol name, etc. So we store that output in a hashset, etc."

Just so that it is eaasier for you (and the others) to modify this if needed when we come back to it in a year.

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+        $start_token = "public symbols"
+        $end_token = "Archive member name"
+        try {
+            [int] $start = $output.IndexOf($start_token)
+            [int] $end = $output.IndexOf($end_token, $start)
+            $symbol_text = ($output.Substring($start + $start_token.Length+1, $end-$start-$start_token.Length-1))
+            $split_symbols = $symbol_text.Split('',[StringSplitOptions]::RemoveEmptyEntries)
+        }
+        catch {
+            Write-Error  "Unable to parse dumpbin output"
```
dcristoloveanu, line 23: >se dumpbin output" [](http://example.com/codeflow?start=39&length=18)

All of these should probably print the lib

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+                    $real_symbol = $split_symbols[$i]
+                    $symbol = $split_symbols[$i].Substring($real_prefix.Length)
+                } else {
+                    $symbol = $split_symbols[$i]
+                    $real_symbol = $real_prefix + $symbol
+                }
+                if(-not ($symbol_table.Add($symbol) -and $symbol_table.Add($real_symbol))){
+                    $return_code = 1
+                    [void]$duplicate_symbols.Add($symbol)
+                }
```
dcristoloveanu, line 48: So we're checking if the "real" and original are present by adding them both to the hashset.
And if the hashset add fails it means that one of them was already added I assume.

dcristoloveanu: At LEAST we should comment that we're doing it this way.

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+        if($duplicate_symbols.Count -gt 0){
+            Write-Error ("The following symbols are duplicate:`n"+($duplicate_symbols -join "`n"))
+        }
```
dcristoloveanu, line 53: Cheers for printing them

dcristoloveanu: We could print them as we find them though. There is no need to insert them and go again throguh the "duplicated symbols".

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+                if($split_symbols[$i].StartsWith($real_prefix)) {
+                    $real_symbol = $split_symbols[$i]
+                    $symbol = $split_symbols[$i].Substring($real_prefix.Length)
+                } else {
+                    $symbol = $split_symbols[$i]
+                    $real_symbol = $real_prefix + $symbol
+                }
+                if(-not ($symbol_table.Add($symbol) -and $symbol_table.Add($real_symbol))){
+                    $return_code = 1
+                    [void]$duplicate_symbols.Add($symbol)
```
dcristoloveanu, line 47: >$duplicate_symbols [](http://example.com/codeflow?start=26&length=18)

To be fair duplicate symbols in the linker land means somthing else usually. It means that 2 object files have the same symbol and at link time the linker finds a duplicate symbol.

But giventhe fact that we can print directly here the variable $duplicate_symbols is not needed so it can be removed.

dcristoloveanu: And I would print something like:
"Original symbol and real symbol found: "

real_check/real_check.ps1
```diff
@@ -0,0 +1,57 @@
...
+            return 2
+        }
+        
+        $real_prefix = "real_"
+        $symbol_table = New-Object -TypeName "System.Collections.Generic.HashSet[string]"
+        $duplicate_symbols = New-Object -TypeName "System.Collections.ArrayList"
+    
+        $return_code = 0
+        for($i=0; $i -lt $split_symbols.Length; $i++){
+            if($i % 2 -ne 0) {
```
dcristoloveanu, line 37: >if($i % 2 -ne 0) { [](http://example.com/codeflow?start=12&length=18)

Is this modulo 2 because of how the formatting of the output of dumpbin is?
In that case that part with adding a comment with an example would actually help.

real_check/test/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+add_subdirectory(reals)
+add_subdirectory(both)
+set(REAL_CHECK_TEST_INC_FOLDER ${CMAKE_CURRENT_LIST_DIR}/inc CACHE INTERNAL "this is what needs to be included if using sharedLib lib" FORCE)
```
dcristoloveanu, line 6: >"this is what needs to be included if using sharedLib lib" [](http://example.com/codeflow?start=76&length=58)

Clearly copy pasted from somewhere else. Nothing with sharedLib in this file.

dcristoloveanu, line 6: >set(REAL_CHECK_TEST_INC_FOLDER ${CMAKE_CURRENT_LIST_DIR}/inc C [](http://example.com/codeflow?start=0&length=62)

Do we need this as a cmake variable? I think not.
Having cmake variables defined like this is useful when the variable is used by other projects.

dcristoloveanu: Probably here you simply want an include_directories?

real_check/test/both/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+set(real_check_test_both_c_files
```
dcristoloveanu, line 3: Can we always have a newline after the copyright header?
Easier to read (this way I have to glance twice on it to figure out where the "code" starts and where the comment header ends.

dcristoloveanu: Here and in all other files

dcristoloveanu:
> 

This file is not needed.
If you wanted ot have it just to look like the typical reals files that we create it is fine to leave it in

But this library with reals could simply have 2 functions

dummy_symbol_1 and dummy_symbol_2

---
Refers to: real_check/test/reals/real_dummy_renames.h:6 in c90fafa. [](commit_id = c90fafa2163891c808426f9ddc142bdbe234f9f6, deletion_comment = False)

real_check/test/src/dummy.c
```diff
@@ -0,0 +1,9 @@
+#include "real_check_test/dummy.h"
```
dcristoloveanu, line 1: >ummy.h [](http://example.com/codeflow?start=27&length=6)

Missing copyright header

real_check/real_check.ps1
dcristoloveanu, line 3: >ction Check-Symbols [](http://example.com/codeflow?start=3&length=19)

I think we should have some tests.
Like we have for the rest of the code.

So I would write the following tests:
- Execute the tool for real_check_test and see that it succeeds
- Execute the tool for real_check_test_reals and see it succeeds
- Execute the tool for both and see it fails

These tests should be automated (a ps1 also) that we should run with cmake (you can add a test to the cmake list of tests and the command for that test be the invocation of powershell with the tests that you add).

dcristoloveanu:
Hah! I wrote my comment and then I saw you said the same :-)

---
In reply to: [932577525](https://github.com/Azure/c-build-tools/pull/80#issuecomment-932577525) [](http://example.com/codeflow?ancestors=932577525,932577525)

dcristoloveanu:
I shall wait for iteration 2 with the addressed comments:
- Would like to see how the code looks like after adding the comments and removing the extra duplicate hash table
- Would like to see the tests for it

dcristoloveanu (changes requested):
:clock1:

parth21999:
> 

Yeah I understand it's not strictly needed. I just created it so the project looks like other ones.

---
In reply to: [932684008](https://github.com/Azure/c-build-tools/pull/80#issuecomment-932684008) [](http://example.com/codeflow?ancestors=932684008,932684008)

---
Refers to: real_check/test/reals/real_dummy_renames.h:6 in c90fafa. [](commit_id = c90fafa2163891c808426f9ddc142bdbe234f9f6, deletion_comment = False)

real_check/test/CMakeLists.txt
parth21999: Removed this line.

parth21999: I agree we don't need the cmake variable. I've removed it in the next iteration. The project seems to build without an `include_directories`.

real_check/test/both/CMakeLists.txt
parth21999: Yes, done in next iteration.

real_check/real_check.ps1
parth21999: Done in next iteration.

real_check/real_check.ps1
parth21999: Done in next iteration.

real_check/real_check.ps1
parth21999: I considered printing as we find them, but decided against doing that. Write-Error generates some formatting for the error message. If we print as we find, the formatting text will be generated for each symbol. This would make the error output hard to read. By collecting and printing once, all the symbols are printed together and the format text is printed only once.

real_check/real_check.ps1
parth21999: Comments added in next iteration.

real_check/real_check.ps1
parth21999: Agreed, adding comments in next iteration.

real_check/real_check.ps1
parth21999: Updated the error message.

real_check/real_check.ps1
parth21999: Added comment in next iteration.

real_check/real_check.md
parth21999: Done.

parth21999: Done.

real_check/CMakeLists.txt
parth21999: Done.

real_check/real_check.ps1
parth21999: Adding test in next iteration.

real_check/real_check.ps1
dcristoloveanu: I see, makes sense

real_check/real_check.ps1
dcristoloveanu: I love that you added the comment with the examples of using the .ps1.
What I meant though is let's add a comment here at line 48 (approx), stating why we are writing the code as such.

Think about it like this:
In 6 months if you come accross this will you remember what the output format of dumpbin is? Wouldn't it be nice to have a glimpse of why you chose to process every other entry?

dcristoloveanu (approved):
:shipit:

PR #81, opened 2021-10-05
add guard for real_check test

parth21999:
Adding guard around test for real_check tool to prevent test from being built by default.


real_check/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 #Copyright (c) Microsoft. All rights reserved.
 #Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-add_subdirectory(test)
+if(${run_unittests})
+    add_subdirectory(test)
+endif()
```
dcristoloveanu, line 6: Let's add a new line at the end of file.

dcristoloveanu (approved):

real_check/CMakeLists.txt
parth21999: Done.

PR #86, opened 2021-10-12
Reals check

parth21999:
Add cmake functionality to collect libs that must be real checked.


build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,22 @@ function(compileTargetAsC11 theTarget)
     set_target_properties(${theTarget} PROPERTIES CXX_STANDARD 11)
   endif()
 endfunction()
+
+set(LIBS_TO_BE_REAL_CHECKED CACHE INTERNAL "LIBS_TO_BE_CHECKED")
+
+function (add_library())
+    _add_library(${ARGV})
+    get_target_property(lib_type ${ARGV0} TYPE)
+    if( (NOT ${CMAKE_CURRENT_LIST_DIR} MATCHES "deps") AND # ignore dependency libs
```
anporumb, line 202: >(NOT ${CMAKE_CURRENT_LIST_DIR} MATCHES "deps") [](http://example.com/codeflow?start=8&length=46)

MATCHES works on regular expressions. If the intention was to exclude everything coming from "deps" then maybe it should have some sort of .* before/after "deps". Like ".*deps.*". Otherwise I am afraid this will exclude only CMakeLists.txt in the root of "deps" folder (if we even have one there).

Same for next line: why would it exclude AAA_ut_lib since AAA_ut_lib if not a match for _ut_lib. It would be a match for .*_ut_lib. 

... not an expert in regex

./runs for the hills.

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,22 @@ function(compileTargetAsC11 theTarget)
...
+        (NOT ${ARGV0} MATCHES "_ut_lib|int_lib$") AND # ignore test libs
```
anporumb, line 203: >$ [](http://example.com/codeflow?start=46&length=1)

the $ will bind to _ut_lib too?

real_check/CMakeLists.txt
```diff
@@ -4,3 +4,5 @@
 if(${run_unittests})
     add_subdirectory(test)
 endif()
+
+add_custom_target(realcheck SOURCES real_check.ps1)
```
anporumb, line 8: Besides having this target in Visual Studio, is there any other reason to have this line here?

anporumb: Also the script should only run on Windows (presumably Linux would produce the same results).

So somewhere there should be some "if LINUX... bye-bye".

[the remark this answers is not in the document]
anporumb: Its documentation says "To add real check to the devops pipeline, add the following step to the `.yml` file and replace `PATH_TO_LIB` and `PATH_TO_REALS_LIB` with appropriate values:"

But I think that now all libraries are collected by the means of overloading add_library.

real_check/real_check.ps1
```diff
@@ -70,7 +70,14 @@ function Check-Symbols {
         return $return_code
 }
 
-exit (Check-Symbols -lib $args[0])
+ForEach($lib in $args){
+    Write-Host "Checking lib $lib"
+    $exit_code = Check-Symbols -lib $lib
```
anporumb, line 75: >$exit_code [](http://example.com/codeflow?start=4&length=10)

what's the exit code from `Check-Symbols` when it throws? I suppose line 45 could thrown some Not Enough Memory exception which doesn't seem to be caught.

real_check/CMakeLists.txt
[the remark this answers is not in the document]
anporumb: I think there's some add_custom_target missing somwhere which only exists when `${run_reals_check}` is true?

real_check/real_check.ps1
anporumb, line 75: >-lib [](http://example.com/codeflow?start=31&length=4)

<not a powershell wizard>, but Check-Symbols doesn't seem to have a named parameter called "lib". </not a powershell wizard>

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,22 @@ function(compileTargetAsC11 theTarget)
...
+        (NOT ${lib_type} STREQUAL "INTERFACE_LIBRARY")) # ignore interface libs
+        set(LIBS_TO_BE_REAL_CHECKED "${LIBS_TO_BE_REAL_CHECKED};${ARGV0}" CACHE INTERNAL "LIBS_TO_BE_CHECKED")
+    endif()
+endfunction()
+
+function (get_target_files())
+    list(SUBLIST ARGV 1 -1 libraries)
```
anporumb, line 210: > list(SUBLIST ARGV 1 -1 libraries) [](http://example.com/codeflow?start=3&length=34)

Why use here list/sublist and then foreach. I mean, what happens if 
```
foreach(lib ${ARGV}) 
```
?

What's the meaning of ARGV0 here and why is it skipped by sublist?

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,22 @@ function(compileTargetAsC11 theTarget)
...
+    foreach(lib ${libraries})
+        set(${ARGV0} "${${ARGV0}} \$\<TARGET_FILE:${lib}\>" CACHE INTERNAL "")
```
anporumb, line 212: >CACHE INTERNAL [](http://example.com/codeflow?start=60&length=14)

There's some funny mechanics with CACHE INTERNAL. Because once it is set, it might remain set between different runs of cmake. And that might not be wanted when a developer is adding let's say a new target to the project.

So I am not sure what's the intention here. Will it all be revealed when the custom target gets added?

I think the review needs the custom target before it can be checked in. Reason to wait.

anporumb (changes requested):
:clock1:

build_functions/CMakeLists.txt
dcristoloveanu, line 202: >"deps" [](http://example.com/codeflow?start=47&length=6)

So if I have a lib named deps_blah.lib this will trigger. Can we make the matching better please?

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,22 @@ function(compileTargetAsC11 theTarget)
     set_target_properties(${theTarget} PROPERTIES CXX_STANDARD 11)
   endif()
 endfunction()
+
+set(LIBS_TO_BE_REAL_CHECKED CACHE INTERNAL "LIBS_TO_BE_CHECKED")
+
+function (add_library())
```
dcristoloveanu, line 199: >add_library [](http://example.com/codeflow?start=10&length=11)

Challenge: can you test this by checking that the libs we expect are there in c-build-tools?

As in if the LIBS_TO_BE_REAL_CHECKED does not contains what we expect for c-build-tools, then let's have an error in cmake.

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,22 @@ function(compileTargetAsC11 theTarget)
...
+        set(LIBS_TO_BE_REAL_CHECKED "${LIBS_TO_BE_REAL_CHECKED};${ARGV0}" CACHE INTERNAL "LIBS_TO_BE_CHECKED")
+    endif()
+endfunction()
+
+function (get_target_files())
+    list(SUBLIST ARGV 1 -1 libraries)
+    foreach(lib ${libraries})
+        set(${ARGV0} "${${ARGV0}} \$\<TARGET_FILE:${lib}\>" CACHE INTERNAL "")
+    endforeach()
+endfunction()
```
dcristoloveanu, line 214: Would be nice to have some comments of what we are doing here. Frankly I would have a hard time changing this code as is.

real_check/CMakeLists.txt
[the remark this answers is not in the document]
dcristoloveanu: I think we won't need a separate step in the yml now, this should simply be picked up by building the solution with the cmake option on, so the .yml stuff can be removed IMO.

dcristoloveanu:
Let's see the next iteration with the yml stuff removed and a bit more comments.

real_check/real_check.ps1
parth21999: This blog seems to suggest that passing the -lib flag is optional when a function parameter is declared using params(). 

[https://devblogs.microsoft.com/scripting/named-arguments-for-powershell-functions-best-practices/](https://devblogs.microsoft.com/scripting/named-arguments-for-powershell-functions-best-practices/)

Might be best to keep the flag for clarity.

parth21999:
> Returns exit code 2 if unable to determine. Prints reason to stderror.

Thanks for pointing that out! Updated in next iteration. Also updated the .md.

---
In reply to: [941822628](https://github.com/Azure/c-build-tools/pull/86#issuecomment-941822628) [](http://example.com/codeflow?ancestors=941822628,941822628)

---
Refers to: real_check/real_check.ps1:105 in 8f26270. [](commit_id = 8f262701ae448d3235270f107f8480333af48eea, deletion_comment = False)

real_check/CMakeLists.txt
parth21999: I added the target so it would rerun real_check if the script was modified. No other reason.

The check is turned off by default. 

I was thinking we can achieve the OS-specific running of the check by only turning on the 'run_realscheck' option in windows builds in the pipeline. Would we still need a OS check in CMakeLists?

build_functions/CMakeLists.txt
parth21999: The first argument (ARGV0) to the function is a variable that will contain the generator expressions for the paths. The rest of the arguments (list(SUBLIST ARGV 1 -1 libraries)) are the libs for which the paths must be found.

real_check/CMakeLists.txt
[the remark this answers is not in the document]
parth21999: """
I think there's some add_custom_target missing somwhere which only exists when ${run_reals_check} is true?
"""

Guarding target with ${run_reals_check} and asserting that platform is windows. Does that resolve your concern @anporumb ?

parth21999: Adding check for platform in next iteration.

build_functions/CMakeLists.txt
parth21999: I clear the variable in every run.

build_functions/CMakeLists.txt
parth21999: done in next iteration.

build_functions/CMakeLists.txt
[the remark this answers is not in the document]
parth21999: Done in next iteration

reals_check/reals_check.md
```diff
@@ -0,0 +1,26 @@
...
+
+To add reals check to a project, add the following snippet to the root `CMakelists.txt` of the project:
+
+```
+if(${run_reals_check})
+    add_reals_check_target([base_name])
+endif()
+```
+
+Replace `\[base_name\]` with the name of the project. CMake will then create a target `\[base_name\]_reals_check`.
```
anporumb, line 26: >with the name of the project [](http://example.com/codeflow?start=24&length=28)

Can this be done automatically? Would CMAKE_PROJECT_NAME work here?

reals_check/test/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+)
+
+set(reals_check_test_h_files
+    ./inc/reals_check_test/dummy.h
+)
+
+add_library(reals_check_test ${reals_check_test_c_files} ${reals_check_test_h_files})
+target_include_directories(reals_check_test PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/reals_check/test/inc>)
+
+add_test(NAME RUN_REALS_CHECK_TEST COMMAND powershell ${CMAKE_CURRENT_SOURCE_DIR}/test_reals_check.ps1 ${CMAKE_CURRENT_SOURCE_DIR}/../reals_check.ps1 ${CMAKE_BINARY_DIR} $<CONFIG>)
```
anporumb, line 18: I understand this test is testing very targeted at the produced libs here (reals_check_test, reals_check_test_reals and reals_check_test_both).

However, there is no exercise of the code (note: line 33 of root CMakeLists.txt) that gathers all the lib with their paths and automatically runs the tool. 

I believe this can be done maybe in the yml of the project? I would generate it simply with -Drun_reals_check:BOOL=ON

anporumb: This is all looking promising, keep at it! I would like to see last iteration before I sign off.

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,40 @@ function(compileTargetAsC11 theTarget)
...
+
+# adds target to run reals check
+function(add_reals_check_target base_target)
+    set(reals_check_target "${base_target}_reals_check")
+    add_custom_target(${reals_check_target} ALL)
+    add_dependencies(${reals_check_target} reals_check ${LIBS_TO_BE_REAL_CHECKED})
+    get_target_files(lib_paths ${LIBS_TO_BE_REAL_CHECKED})
+    add_custom_command(TARGET ${reals_check_target} POST_BUILD
+        COMMAND
+        powershell.exe ${CMAKE_CURRENT_LIST_DIR}/deps/c-build-tools/reals_check/reals_check.ps1 ${lib_paths}
```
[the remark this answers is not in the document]
parth21999: c-build-tools does not produce any libs that need to be checked so there's no need add the target.

reals_check/reals_check.md
parth21999: Yes that would work. Using CMAKE_PROJECT_NAME in the next iteration.

reals_check/test/inc/reals_check_test/dummy.h
```diff
@@ -0,0 +1,19 @@
...
+
+#ifndef DUMMY_H
+#define DUMMY_H
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+int dummy_symbol_1();
```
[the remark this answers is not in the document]
parth21999: Thanks for catching that. Done in next iteration.

reals_check/test/CMakeLists.txt
parth21999: I've added a test to make sure that the correct libs are being selected for reals checking. 

Also adding the reals check target to run the check automatically.

build/devops.yml
```diff
@@ -142,7 +142,7 @@ jobs:
     displayName: 'CMake ..  -Drun_unittests:bool=ON -Drun_int_tests:bool=ON -Drun_perf_tests:bool=ON -Duse_cppunittest:bool=ON -Drun_e2e_tests:BOOL=ON -Dbuild_traceability_tool:bool=OFF -G "Visual Studio 16 2019" -A x64'
     inputs:
       workingDirectory: 'build_x64'
-      cmakeArgs: '..  -Drun_unittests:bool=ON -Drun_int_tests:bool=ON -Drun_perf_tests:bool=ON -Duse_cppunittest:bool=ON -Drun_e2e_tests:BOOL=ON -Dbuild_traceability_tool:bool=OFF -G "Visual Studio 16 2019" -A x64'
+      cmakeArgs: '.. -Drun_reals_check:bool=ON -Drun_unittests:bool=ON -Drun_int_tests:bool=ON -Drun_perf_tests:bool=ON -Duse_cppunittest:bool=ON -Drun_e2e_tests:BOOL=ON -Dbuild_traceability_tool:bool=OFF -G "Visual Studio 16 2019" -A x64'
```
anporumb, line 145: >-Drun_reals_check:bool=ON [](http://example.com/codeflow?start=21&length=25)

not needed for Semmle, remove.

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,43 @@ function(compileTargetAsC11 theTarget)
...
+
+# variable to store list of libs that must be checked for reals
+set(LIBS_TO_BE_REAL_CHECKED CACHE INTERNAL "LIBS_TO_BE_CHECKED")
+
+# overloading the add_library function to store name of library if it meets criteria
+function (add_library())
+    _add_library(${ARGV})
+    get_target_property(lib_type ${ARGV0} TYPE)
+    if( (NOT ${CMAKE_CURRENT_LIST_DIR} MATCHES "/deps/") AND # ignore dependency libs
+        (NOT ${ARGV0} MATCHES "_ut_|_int_|_perf_") AND # ignore test libs
```
anporumb, line 205: >_perf_ [](http://example.com/codeflow?start=42&length=6)

frankly, perf should not be here.

Perf is always run against the production code (or it should be). If "real" symbols sneak in - mistake then and should be fixed by removing the "real" symbols.

build_functions/CMakeLists.txt
```diff
@@ -193,3 +193,43 @@ function(compileTargetAsC11 theTarget)
     set_target_properties(${theTarget} PROPERTIES CXX_STANDARD 11)
   endif()
 endfunction()
+
+# variable to store list of libs that must be checked for reals
+set(LIBS_TO_BE_REAL_CHECKED CACHE INTERNAL "LIBS_TO_BE_CHECKED")
+
+# overloading the add_library function to store name of library if it meets criteria
+function (add_library())
+    _add_library(${ARGV})
```
anporumb, line 202: >_add_library [](http://example.com/codeflow?start=4&length=12)

this overload is not a supported feature of CMake. Would be nice to have a link to the original place where you've picked up this trick

reals_check/test/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+# Adding libs that should get ignored for reals check
+add_library(random_int_test ./dummy_source.c)
+add_library(random_ut_test ./dummy_source.c)
+add_library(random_perf_test ./dummy_source.c)
+add_library(random_interface_lib INTERFACE ./dummy_source.c)
```
anporumb, line 24: >dummy_source [](http://example.com/codeflow?start=45&length=12)

funny seeing a .c (source) file as a constituent of a INTERFACE library. Maybe you mean to have a header here?

reals_check/test/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    ./inc/reals_check_test/dummy.h
+)
+
+add_library(reals_check_test ${reals_check_test_c_files} ${reals_check_test_h_files})
+target_include_directories(reals_check_test PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/reals_check/test/inc>)
+
+add_test(NAME RUN_REALS_CHECK_TEST COMMAND powershell ${CMAKE_CURRENT_SOURCE_DIR}/test_reals_check.ps1 ${CMAKE_CURRENT_SOURCE_DIR}/../reals_check.ps1 ${CMAKE_BINARY_DIR} $<CONFIG>)
+
+# Adding libs that should get ignored for reals check
+add_library(random_int_test ./dummy_source.c)
```
anporumb, line 21: >random_int_test [](http://example.com/codeflow?start=12&length=15)

the names of these libraries should match the names we usually have in the rest of our repos.

Here's some examples:

sm_ut_lib, strings_ut_lib, sm_int_lib. Note how these names have _ut_lib or _int_lib as their endings.

This change should go to _add_library overload too.

reals_check/test/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+add_library(random_ut_test ./dummy_source.c)
+add_library(random_perf_test ./dummy_source.c)
+add_library(random_interface_lib INTERFACE ./dummy_source.c)
+
+set(expected_value ";reals_check_test_reals;reals_check_test")
+
+# Test to check that the correct libraries are selected for reals checking:
+if(NOT "${LIBS_TO_BE_REAL_CHECKED}" STREQUAL "${expected_value}")
+    message(FATAL_ERROR "LIBS_TO_BE_REAL_CHECKED does not contain the correct value.\nExpected value: ${expected_value}\nActual value: ${LIBS_TO_BE_REAL_CHECKED}")
```
anporumb, line 30: > ${expected_value}\nActual value: ${LIBS_TO_BE_REAL_CHECKED}") [](http://example.com/codeflow?start=101&length=62)

This is nice

build_functions/CMakeLists.txt
dcristoloveanu: Nice, I saw the test

build_functions/CMakeLists.txt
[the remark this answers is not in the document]
parth21999: I'm not sure I understand your concern. Adding "_int_" and "_perf_" make sure that those libs are not checked for reals. Since real checking happens in jobs where int tests are built, we need "_int_" here to exclude those libs from getting checked.

PR #88, opened 2021-10-19
fix reals check target

parth21999:


build_functions/CMakeLists.txt
```diff
@@ -227,6 +227,9 @@ endfunction()
 # adds ${CMAKE_PROJECT_NAME}_reals_check as target to run reals check
 # signature: add_reals_check_target()
 function(add_reals_check_target)
+    if(NOT WIN32)
+        message(FATAL_ERROR "Cannot run reals check for non-Windows build.")
```
anporumb, line 231: >FATAL_ERROR [](http://example.com/codeflow?start=16&length=11)

So if add_reals_check_target is called for Linux it would error out, right?

That means that every Linux build needs to have if(WIN32)... add target ... endif().

How about "if it is not win32 then silently do nothing?"

That way nobody needs to care about the OS

parth21999: add_reals_check_target is only called if `run_reals_check` is set to `ON`, which should only be done for windows builds.

PR #89, opened 2021-10-20
Fix regex to ignore libs for reals check

parth21999:


build_functions/CMakeLists.txt
```diff
@@ -203,7 +203,7 @@ function (add_library())
     _add_library(${ARGV})
     get_target_property(lib_type ${ARGV0} TYPE)
     if( (NOT ${CMAKE_CURRENT_LIST_DIR} MATCHES "/deps/") AND # ignore dependency libs
-        (NOT ${ARGV0} MATCHES "_(ut|int|perf)_(lib|dll)") AND # ignore test libs
+        (NOT ${ARGV0} MATCHES "_(ut|int|perf)_lib|dll") AND # ignore test libs and dlls
```
dcristoloveanu, line 206: >_(ut|int|perf)_lib|dll" [](http://example.com/codeflow?start=31&length=23)

regular expressions manage to confuse me all the time.
I was looking at MATCHES ([https://cmake.org/cmake/help/latest/command/if.html](https://cmake.org/cmake/help/latest/command/if.html))
And it says it uses this Regex specs: [https://cmake.org/cmake/help/latest/command/string.html#regex-specification](https://cmake.org/cmake/help/latest/command/string.html#regex-specification)

And I see that () is for saving the matched expression
Sooo, how does this change fix anything? And what?

dcristoloveanu: Consider stating it like : "test libs and any other dlls"

PR #100, opened 2022-01-13
Fix script

parth21999:
Changed script to use loop instead of recursion to avoid call depth overflow issue.


jebrando (approved):
:shipit:

anporumb (approved):
:shipit:

PR #103, opened 2022-01-15
refresh deps

parth21999:
Remove deps folder to refresh deps.

For some reason, sometimes submodule updates don't get detected. Deleting the deps folder fixes that problem somehow.

### DIFF
--- a/update_deps/propagate_updates.ps1
+++ b/update_deps/propagate_updates.ps1
@@ -98,6 +98,8 @@ function update-local-repo {
     cd $repo_name
     git checkout master
     git pull
+    # Remove deps folder to refresh
+    Remove-Item deps -Recurse -Force
     git submodule update --init
     # update all submodules except the ones mentioned in ignores.json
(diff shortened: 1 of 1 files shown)


update_deps/propagate_updates.ps1
```diff
@@ -98,6 +98,8 @@ function update-local-repo {
     cd $repo_name
     git checkout master
     git pull
+    # Remove deps folder to refresh
+    Remove-Item deps -Recurse -Force
```
anporumb, line 102: I don't understand why this is needed.

But if e are in paranoid state, why not run a git clean -xdff too while we are at it.

anporumb (approved):
:shipit:

update_deps/propagate_updates.ps1
parth21999: Even I don't understand why dependency updates aren't being updated. This fixes it somehow.

PR #104, opened 2022-01-24
Update traceabilitytool to use .NET 4.8 so win11/vs22 works

mattdurak:
I couldn't load this project in a clean win11/vs2022 setup otherwise, and .NET 4.5 doesn't seem to be available anymore.


dcristoloveanu (approved):
:shipit:

PR #105, opened 2022-01-26
add sleep

parth21999:
Adding brief sleep to allow mergeable state to stabilize after all checks have passed.


mattdurak:
>                 }

If this is "unstable" should we loop/retry (at least up to some max retry count)?

Then you might not need the sleep except in the retry case.

---
Refers to: update_deps/propagate_updates.ps1:206 in c8e0943. [](commit_id = c8e094365ca28da4649314b4dd0986026ec5b00c, deletion_comment = False)

mattdurak (approved):
:shipit:

PR #106, opened 2022-02-11
Build with VS2022

dcristoloveanu:


dcristoloveanu:
This change to build with VS2022 has already been merged with another PR.

PR #110, opened 2022-04-23
only delete deps that are submodules

parth21999:


update_deps/propagate_updates.ps1
```diff
@@ -87,6 +87,16 @@ create-ignore-pattern
 # name for branch which will be used to create PR
 $new_branch_name = "new_deps"
 
+function refresh-submodules {
+    $submodules = git submodule | Out-String
```
jebrando, line 91: What would be really nice to see if the user has any changes in those repositories and warn them that their changes will be removed.  Just something to think about.

jebrando (approved):

update_deps/propagate_updates.ps1
parth21999: The best way to update dependencies is to run the script in a clean directory. That way all the repos will be cloned in that directory from scratch and no work will be affected.

PR #117, opened 2022-07-22
Add Hari to FabricBot

dcristoloveanu:


darobs (approved):

.github/fabricbot.json
```diff
@@ -128,6 +128,12 @@
             "parameters": {
               "reviewer": "ACKPaul"
             }
+          },
+          {
+            "name": "haridamineni",
```
jebrando, line 133: >"name": "haridamineni" [](http://example.com/codeflow?start=12&length=22)

Shouldn't this be requestReviewer?

jebrando (approved):
:shipit:

.github/fabricbot.json
dcristoloveanu: Wow, thanks for catching that!

PR #122, opened 2022-08-29
fsanitize_address is now an option for Visual Studio

anporumb:


CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(run_traceability "run traceability tool (default is ON)" ON)
 option(build_traceability_tool "build traceability tool (default is ON)" ON)
 option(run_reals_check "run reals check tool (default is OFF)" OFF)
 option(use_vld "use VLD for leak detection (default is OFF)" OFF)
+option(fsanitize_address "enable /fsanitize=address compiler switch (default is OFF)" OFF) # see https://docs.microsoft.com/en-us/cpp/build/reference/fsanitize
```
dcristoloveanu, line 35: >fsanitize_address [](http://example.com/codeflow?start=7&length=17)

I see the downstream repos (like ebs) has fsanitize as CmakeLists option.

Seems like maybe it was actually missed in this repo originally?
Can we keep only one option (I will get so confused if we end up having fsanitize and fsanitize_address as cmakelists options ...)

dcristoloveanu: Pretty please!

dcristoloveanu: With much many sprinkles!

PR #125, opened 2022-09-23
add /CETCOMPAT to linker

anporumb:


build_functions/CMakeLists.txt
```diff
@@ -133,13 +133,13 @@ macro(set_default_build_options)
         endif()
 
         # /guard:cf is Enable Control Flow Guard https://docs.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard, working for compiler and the linker
-        set(CMAKE_EXE_LINKER_FLAGS "/INCREMENTAL:NO ${CMAKE_EXE_LINKER_FLAGS} /LTCG /IGNORE:4075 /WX /guard:cf")
-        set(CMAKE_SHARED_LINKER_FLAGS "/INCREMENTAL:NO ${CMAKE_SHARED_LINKER_FLAGS} /LTCG /IGNORE:4075 /WX /guard:cf")
+        # /CETCOMPAT (CET Shadow Stack compatible) https://learn.microsoft.com/en-us/cpp/build/reference/cetcompat. Also mitigates: warning BA2025: 'latency_test_server.exe' does not enable the Control-flow Enforcement Technology (CET) Shadow Stack mitigation. To resolve this issue, pass /CETCOMPAT on the linker command lines.
+        set(CMAKE_EXE_LINKER_FLAGS "/INCREMENTAL:NO ${CMAKE_EXE_LINKER_FLAGS} /LTCG /IGNORE:4075 /WX /guard:cf /CETCOMPAT")
```
dcristoloveanu, line 137: >/CETCOMPAT [](http://example.com/codeflow?start=111&length=10)

Does this work with vld on? VLD does do some hijacking, so we might need to enable this for non-vld builds only.

dcristoloveanu: Would hate that we propagate all the way up and then find out our build does not pass, so I think it would be good to have a run of the main repo with this changes first.

dcristoloveanu (approved):
:shipit:

build_functions/CMakeLists.txt
anporumb: I would suspect this works with ALL switches, because apparently it is a processor (hardware) feature from Intel.

`Control-flow Enforcement Technology (CET) Shadow Stack is a computer processor feature. `

But nevertheless, I will make a run of the ebs repo with this switch turned on BEFORE I check this in.

anporumb: [https://msazure.visualstudio.com/One/_git/Azure-MessagingStore/pullrequest/6800093](https://msazure.visualstudio.com/One/_git/Azure-MessagingStore/pullrequest/6800093) (if it finds the SHAs) will tell if it works.

anporumb: I don't see any conflicts so far with the build, and I am all out of patiuence. Will check it in.

PR #126, opened 2022-09-26
Improve propagate_updates.ps1

parth21999:
These changes address the following issues:

- The script would accept a file path as `-root`. This is undesirable because the script would still work in most cases, unless the remotes for some repositories were set to not be Azure or Github. These changes validate that the value passed to the `-root` argument is a valid URL.
- `-azure_work_item` was optional. If `-azure_work_item` was not provided, the script would fail when it would try to update an Azure repository. These changes make `-azure_work_item` mandatory. This allows the script to fail early if the `-azure_work_item` is not provided. This has the implication that the script cannot be used to only propagate updates to a repository in the middle of dependency graph. Since that is not a common use case, this change is justified.
- `-root` was optional and assumed to be `https://msazure.visualstudio.com/DefaultCollection/One/_git/Azure-MessagingStore`. Now it is mandatory since the script is being used to propagate updates to other top-level projects (ElasticLog, Georeplication).

### DIFF
--- a/update_deps/build_graph.ps1
+++ b/update_deps/build_graph.ps1
@@ -51,6 +51,11 @@ function get-name-from-url {
     param (
         [string] $url
    )
+   if(!$url.Contains("http"))
+   {
+        Write-Error("Invalid URL: $url")
+        exit -1
+   }
    $split_by_slash = $url.Split('/')
    $split_by_dot = $split_by_slash[-1].Split('.') # $split_by_slash[-1] contains [repo_name].git
    return $split_by_dot[0] # $split_by_dot[0] contains [repo_name]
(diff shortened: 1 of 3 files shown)


update_deps/build_graph.ps1
```diff
@@ -51,6 +51,11 @@ function get-name-from-url {
     param (
         [string] $url
    )
+   if(!$url.Contains("http"))
```
dcristoloveanu, line 54: I wonder why this was needed. It should not be needed, right?

parth21999: From the PR description:

"The script would accept a file path as -root. This is undesirable because the script would still work in most cases, unless the remotes for some repositories were set to not be Azure or Github. These changes validate that the value passed to the -root argument is a valid URL. "

dcristoloveanu (approved):
:shipit:

PR #128, opened 2023-01-05
Add vld.ini file which contains ignore functions list 

cyrus-jackson:
Adds these methods 
1. `anonymous namespace'::GetOSVersion,
2. Azure::Storage::_internal::UrlEncodeQueryParameter,
3. Azure::Storage::_internal::UrlEncodePath,Azure::Core::Url::Decode,
4. Azure::Identity::_detail::TokenCredentialImpl::GetToken

to `IgnoreFunctionsList` in vld.ini file. 

These methods are from sdk and because of these methods our vld reports them as leaks due to static string allocation.

### DIFF
--- a/build_functions/CMakeLists.txt
+++ b/build_functions/CMakeLists.txt
@@ -160,6 +160,8 @@ macro(set_default_build_options)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /fsanitize=address")
     endif()
 
+    configure_file(${build_c_tests_internal_dir}/ignore_function_names_vld.ini $ENV{ProgramFiles\(x86\)}/Visual Leak Detector/vld.ini COPYONLY)
+    
     enable_testing()
 endmacro()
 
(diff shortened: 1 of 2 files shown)


build_functions/CMakeLists.txt
```diff
@@ -160,6 +160,8 @@ macro(set_default_build_options)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /fsanitize=address")
     endif()
 
+    configure_file("${build_c_tests_internal_dir}/ignore_function_names_vld.ini" "$ENV{ProgramFiles\(x86\)}/Visual Leak Detector/vld.ini" COPYONLY)
```
dcristoloveanu, line 163: > configure_file("${build_c_tests_internal_dir}/ignore_function_names_vld.ini" "$ENV{ProgramFiles\(x86\)}/Visual Leak Detector/vld.ini" COPYONLY) [](http://example.com/codeflow?start=3&length=144)

Shouldn't this just come with the new vld version?
I think we should not mangle existing versions (copying to program files) with our SW.

PR #130, opened 2023-01-24
Add another function to install headers only by default.

darobs:
Add another function to install headers only by default.


mattdurak (approved):
:shipit:

PR #141, opened 2023-06-13
Remove add_library override

mattdurak:
vcpkg uses an add_library override and it's not officially supported. Since we also had one we could not use the vcpkg cmake functions. But we don't need that override, we just need a way to collect a list of libraries that need to have the reals check run on. This can be done without hooking add_library.


dcristoloveanu (approved):
:shipit:

PR #144, opened 2023-08-04
Add devops yaml templates for use in other repos

mattdurak:
Moved some of the common yaml from the EBS/ELS projects here for re-use:
- ETL logging
- dumping large file list
- Moved the tttracer yml with these templates because they can be used directly without copying


anporumb (approved):
:shipit:

PR #146, opened 2023-09-06
adding -fPic flag to linux build

jebrando:


darobs (approved):
TIL a little bit more about PIC.

PR #151, opened 2023-10-14
Make a build_as_csharp_net6_windows_project function

darobs:
Make a build_as_csharp_net6_windows_project function to use when we require Windows Framework.


avranju (approved):
🚢

PR #160, opened 2023-11-02
Add support to SRS extension to strip SRS tags

mattdurak:
Can be helpful if copying requirements from another file and re-tagging them with a different prefix. Just does a simple regex for now but could be made smarter.
Also restrict keybindings to only work when there is text selected.
extension version 0.0.9


jebrando (approved):
:shipit:

PR #163, opened 2023-11-09
Propagate Dependencies Updates Script Improvements

parth21999:
- Updates are committed on a branch with a unique name
 - Changed script to use GitHub CLI instead of REST API which was unreliable


update_deps/propagate_updates.ps1
```diff
@@ -14,12 +14,6 @@ root repo by making PRs to each repo in bottom-up level-order.
 
 .PARAMETER root
 
-URL of the repository upto which updates must be propagated.
```
mattdurak, line 17: shouldn't be deleted I assume

update_deps/propagate_updates.ps1
```diff
@@ -116,151 +107,38 @@ function update-local-repo {
...
+        [string] $repo_name,
+        [string] $new_branch_name
     )
+    cd $repo_name
     Write-Host "`nCreating PR"
-    $api_url, $html_url = create-pr-github $repo_name
+    $working_directory = (Get-Location).Path
+    gh pr create --title "[autogenerated] update dependencies" --body "Propagating dependency updates" --head $new_branch_name
+    Write-Host "Waiting for checks to start"
+    Start-Sleep -Seconds 30
```
mattdurak, line 130: >Start-Sleep -Seconds 30  [](http://example.com/codeflow?start=4&length=24)

curious why a 30s wait is needed...but probably doesn't matter because none of our gates likely can run any faster than that...

mattdurak (dismissed):
:shipit:

mattdurak (approved):
:shipit:

update_deps/propagate_updates.ps1
parth21999: It takes a moment for the checks to start. If `gh pr checks --watch` is called before the checks are started, it can skip over the checks that have not been started.

The 30s sleep allows enough time for all the checks to start

update_deps/propagate_updates.ps1
parth21999: Fixed.

PR #176, opened 2024-03-19
Removing the -V option in ctest

jebrando:
The -V option enable all test logging to get shown on the screen.  This option causes log files to be large and doesn't is not necessary.


dcristoloveanu (approved):
:shipit:

PR #177, opened 2024-03-23
Remove the -V option from ctest runs as it is too much logging

dcristoloveanu:
Remove the -V option from ctest runs as it is too much logging


jebrando (approved):
:shipit:

PR #180, opened 2024-03-29
Adjust some defaults and remove todo for x86

dcristoloveanu:


pipeline_templates/build_all_flavors.yml
```diff
@@ -4,7 +4,7 @@ parameters:
   # GBALLOC_LL_TYPE_VALUES is an array of strings
   - name: GBALLOC_LL_TYPE_VALUES
     type: object
-    default: ["PASSTHROUGH", "JEMALLOC"]
+    default: ["PASSTHROUGH"]
```
anporumb, line 7: >default: ["PASSTHROUGH"] [](http://example.com/codeflow?start=4&length=24)

errr... I don't think this should be removed.

anporumb (approved):
:shipit:

pipeline_templates/build_all_flavors.yml
dcristoloveanu: I plan to override it in the repos that use that (c-pal and so forth). macro-utils, ctest, umock-c, c-logging and so on we don't build with JEMALLOC. To be honest this dimension would be nicer as an addon dimension alltogether (as in if one could specify some custom dimensions that we iterate through maybe that would be nicer, but I just don't want to spend my juice on that now - I really REALLY don't want it, not that it would not be nice to do).

PR #183, opened 2024-04-03
Propagate Updates Improvements

parth21999:
- Increased wait time for checks to begin from 30s to 120s. Checks seem to be taking longer than before to start (corporate terrorism) so this helps take that into account. Checks take longer than 120s to run anyway so script doesn't take longer on the whole.
- Print current directory in warning message


mattdurak (approved):
:shipit:

PR #190, opened 2024-05-24
Have an option to link to the Windows CRT statically

mattdurak:
Adds vcpkg triplet and sets the `CMAKE_MSVC_RUNTIME_LIBRARY` based on a new parameter `fstatic_win32_crt` (default OFF)


dcristoloveanu:
Stale PR

PR #194, opened 2024-07-11
Appverif in pipeline template

mattdurak:
Optionally allow app verifier run in build_all_flavors

### DIFF
--- a/pipeline_templates/build_and_run_tests.yml
+++ b/pipeline_templates/build_and_run_tests.yml
@@ -22,6 +22,14 @@ parameters:
   - name: run_test_suite_without_ctest
     type: string
     default: ""
+  # Settings to add an app verifier step for ctest
+  - name: with_app_verifier
+    type: boolean
+    default: false
+  - name: repo_root_override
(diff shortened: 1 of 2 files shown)


pipeline_templates/build_and_run_tests.yml
```diff
@@ -115,6 +123,21 @@ jobs:
         arguments: '-C "${{ parameters.build_configuration }}" --output-on-failure'
         workingFolder: 'cmake_build'
 
+  - ${{ if and(eq(parameters.FSANITIZE_TYPE, 'OFF'), and(eq(parameters.run_test_suite_without_ctest, ''), parameters.with_app_verifier)) }}:
```
mattdurak, line 126: >run_test_suite_without_ctest [](http://example.com/codeflow?start=71&length=28)

note: for now I'm just adding this for the basic case of "run all tests in ctest". It can be expanded later for other `run_test_suite_without_ctest` cases, but I didn't need that for zrpc.

jebrando (approved):
:shipit:

PR #196, opened 2024-08-02
Fix appverifier in the build_all_flavors.yml to allow excluding a test

mattdurak:
Exposes the cmake options to pass "-E" to exclude tests from appverifier


jebrando (approved):
:shipit:

PR #198, opened 2024-08-15
sometimes stop doesnt stop trace

anporumb:


pipeline_templates/scripts/stop_trace.ps1
```diff
@@ -21,8 +21,32 @@ try
...
+
+    # Setup a timer
+    $timer = New-Object System.Timers.Timer
+    $timer.Interval = $timeout
+    $timer.AutoReset = $false
+
+    # Define the action to take when the timer elapses
+    $timerEvent = Register-ObjectEvent -InputObject $timer -EventName Elapsed -Action {
+        Write-Host "Timeout reached. Restarting the machine! Expect the build to (eventually) fail... "
+        Restart-Computer -Force
```
mattdurak, line 38: We have sysinternals on the gate machines, I wonder if we should bugcheck here with NotMyFault and get a dump to send to the Windows team if the ETW logging is stuck? [https://learn.microsoft.com/en-us/sysinternals/downloads/notmyfault](https://learn.microsoft.com/en-us/sysinternals/downloads/notmyfault)

Just an idea...

mattdurak (approved):
:shipit:

pipeline_templates/scripts/stop_trace.ps1
anporumb: We've talked and decided to procdump + kernel crash.

... something will happen :)

mattdurak (approved):
:shipit:

PR #200, opened 2024-08-25
Use vcpkg package caching

dcristoloveanu:


build_functions/CMakeLists.txt
```diff
@@ -336,6 +336,9 @@ macro(use_vcpkg vcpkgPath)
     # Only obtain assets from Terrapin project. See Asset Caching (https://vcpkg.io/en/docs/users/assetcaching.html)
     set(ENV{X_VCPKG_ASSET_SOURCES} "clear;x-azurl,https://vcpkg.storage.devpackages.microsoft.io/artifacts/;x-block-origin")
 
+    #Use vcpkg package caching into an ADO artifact feed
+    set(ENV{VCPKG_BINARY_SOURCES} "clear;nuget,https://msazure.pkgs.visualstudio.com/One/_packaging/azure-messagingstore-vcpkg-cache/nuget/v3/index.json,readwrite")
```
BertKleewein, line 340: >nuget [](http://example.com/codeflow?start=41&length=5)

I'm assuming what the syntax means and wondering if order matters. Does this mean "check the nuget feed first, then use the ADO feed second""?

BertKleewein (approved):
:shipit:

build_functions/CMakeLists.txt
dcristoloveanu: Actually, syntax means use a Nuget server and after the comma the endpoint to use. So there is just one entry to use of where the packages should be read from/written to.
https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-nuget?pivots=shell-cmd

PR #209, opened 2024-09-27
Remove attempt to get Azure VM size - fails all the time

dcristoloveanu:
Remove attempt to get Azure VM size - fails all the time


jebrando (approved):
:shipit:

PR #211, opened 2024-10-03
incremental check of submodules

anporumb:


pipeline_templates/run_master_check.yml
```diff
@@ -1,13 +1,24 @@
...
+# verifies that all submodules are ancestors of their master branch
 
 parameters:
- - name: repo_root
-   default: $(Build.SourcesDirectory)/deps/c-build-tools
+  - name: repo_root
+    default: $(Build.SourcesDirectory)/deps/c-build-tools
 
-steps:
+jobs:
```
BertKleewein, line 7: >jobs [](http://example.com/codeflow?start=0&length=4)

I assume there are other YAML changes in EBS to correctly reference this file, right? Otherwise, I don't see how you can insert this additional level without breaking the YAML that includes this.

BertKleewein (approved):
:shipit:

dcristoloveanu (approved):
:shipit:

pipeline_templates/run_master_check.yml
anporumb: True, what I wanted to achieve and failed in the first iteration was to have parity with `codeql3000_default.yml` which is called "as is" by the pipeline.

What I wrote wasn't "evil", but it wasn't a 1 line replacer in all the other repos. Instead it would have to be used like "disable_appverifier" which is a "something" in another "something bigger".

But checking submodules is separated enough that it can be its own "jobs" collection.

PR #216, opened 2024-11-21
Fix a miss on how we pass the parameter for the codeql3000 default yml

dcristoloveanu:
Fix a miss on how we pass the parameter for the codeql3000 default yml


jebrando (approved):
:shipit:

PR #217, opened 2024-12-06
Add delete branch option to propagate_updates for the azure devops repos

mattdurak:
This should help reduce the branches that get left behind in the ADO repos


mattdurak:
I'm running the script with this option now, will verify that it works first

anporumb (approved):
:shipit:

PR #220, opened 2025-01-16
use jobs instead of timers for logman. Add use_guard_cf. Reorganize a bit compiler/linker switches

anporumb:


jebrando (approved):
:shipit:

PR #233, opened 2025-04-11
Fix to make RelWithDebInfo and MinSizeRel optimized for C#

mattdurak:
Was not optimized by default:
image


dcristoloveanu (approved):
:shipit:

PR #235, opened 2025-05-22
Remove environment variable Build_SourcesDirectory

mattdurak:
We can use the cmake variable `PROJECT_SOURCE_DIR` instead


dcristoloveanu (approved):
:shipit:

PR #238, opened 2025-07-16
Add ignore submodules parameter for master check

dcristoloveanu:
Add ignore submodules parameter for master check in order to allow skipping specific submodules when performing the master check

### DIFF
--- a/pipeline_templates/scripts/submodule_master_check.ps1
+++ b/pipeline_templates/scripts/submodule_master_check.ps1
@@ -54,27 +54,16 @@ foreach ($submodule in $submodules) {
     $ref = "refs/remotes/origin/$branchToCheck"
     $originBranch = "origin/$branchToCheck"
 
-    # Check if the master branch exists
-    git show-ref --verify --quiet $ref
+    # Check if the SHA is an ancestor of the desired branch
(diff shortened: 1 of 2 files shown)


pipeline_templates/scripts/submodule_master_check.ps1
```diff
@@ -2,41 +2,77 @@
...
+
     # Check if the master branch exists
-    git show-ref --verify --quiet refs/remotes/origin/master
+    git show-ref --verify --quiet $ref
 
     # Capture the exit code
     $exitCode = $LASTEXITCODE
 
     if ($exitCode -gt 0) {
-        Write-Host "Master branch not found in $submodulePath, skipping..."
```
BertKleewein, line 28: >Write-Host "Master branch not found in $submodulePath, skipping..." [](http://example.com/codeflow?start=8&length=67)

I think we still need to keep this hack, right? I feel like we'd be better off without it.

dcristoloveanu: I can remove it and see what breaks :-)

BertKleewein (approved):
:shipit:

BertKleewein (approved):
:shipit:

PR #343, opened 2026-03-06
[MrBot] Add build_reals option to gate reals libraries

dcristoloveanu:
## Summary
Add `build_reals` CMake option to gate reals libraries behind a flag.

## Motivation
The `build_reals` option (default ON) allows disabling the building of reals libraries when they are not needed — such as in official builds, integration test builds, and performance test builds. Reals are only consumed by unit tests, so excluding them from non-unit-test configurations reduces build times.

This is part of a broader effort to reduce the Azure-MessagingStore official simulation build time (currently ~86 min).

## Changes
- Add `build_reals` option definition (default ON) alongside other test options in `tests/CMakeLists.txt` behind `if(\${build_reals})`


anporumb (approved):
:shipit:

PR #351, opened 2026-03-13
Remove IWYU (include-what-you-use) files

anporumb:
IWYU is no longer used. Remove:
- build/linux/build_linux_iwyu.sh - IWYU build script
- iwyu/rules.imp - IWYU mapping rules


dcristoloveanu (approved):
:shipit:

PR #380, opened 2026-04-07
[MrBot] Add setup_nuget parameter to run NuGetAuthenticate before cmake

anporumb:
Some repos (e.g. azure-messaging-metrics) download NuGet packages during cmake configure via `execute_process(COMMAND nuget ...)`. On build agents the Azure Artifacts credential provider needs `NuGetAuthenticate@1` to have run first, otherwise `GetTokenSilently` fails with `0x80070520`.

This adds an opt-in `setup_nuget` boolean parameter (default `false`) to both `build_and_run_tests.yml` and `build_all_flavors.yml`. When `true`, the existing `setup_nuget_tools.yml` template is called after submodule checkout and before cmake configure.

**Usage:**
\\\yaml
- template: /pipeline_templates/build_all_flavors.yml@c_build_tools
  parameters:
    setup_nuget: true
\\\

**Changes:**
- `build_and_run_tests.yml`: New `setup_nuget` parameter; conditionally calls `setup_nuget_tools.yml`
- `build_all_flavors.yml`: New `setup_nuget` parameter; passes through to `build_and_run_tests`

No impact on existing consumers (default is `false`).

### DIFF
--- a/pipeline_templates/build_all_flavors.yml
+++ b/pipeline_templates/build_all_flavors.yml
@@ -74,6 +74,12 @@ parameters:
     type: boolean
     default: false
 
+  # When true, run NuGetAuthenticate and NuGetToolInstaller before cmake
+  # so that execute_process(COMMAND nuget ...) can access private feeds.
+  - name: setup_nuget
+    type: boolean
+    default: false
+
 jobs:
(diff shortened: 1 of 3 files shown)


jebrando (approved):
:shipit:

PR #381, opened 2026-04-07
Run setup_nuget_tools.yml before cmake generate

anporumb:
Call setup_nuget_tools.yml (NuGetAuthenticate + NuGetToolInstaller) in the cmake.yml wrapper template so that any execute_process(COMMAND nuget ...) during cmake configure can authenticate to Azure Artifacts feeds.


mattdurak (approved):
:shipit:

PR #389, opened 2026-04-10
Fix C# test file detection for .NET-style naming conventions

anporumb:
The is_test_file() function only recognized C-style test directories (ending with _ut or _int) and C# files starting with 'Test'. This caused 708 false positives in EBS where test files use .NET conventions:
- Directories: *.UnitTests, *.IntTests
- Filenames: *Tests.cs

Now also matches directory names ending with UnitTests or IntTests and filenames ending with Tests.cs.

Added test fixture cs_clean_dotnet_naming to cover the new patterns.


dcristoloveanu (approved):
:shipit:

PR #391, opened 2026-04-10
[MrBot]  Add singular Test.cs pattern to is_test_file check

BertKleewein:
The is_test_file function now also recognizes C# filenames ending with 'Test.cs' (singular) in addition to 'Tests.cs' (plural) and filenames starting with 'Test'.


anporumb (approved):
:shipit:

PR #393, opened 2026-04-22
[MrBot] Suppress CMake regeneration to fix MSB3371 ZERO_CHECK races

anporumb:
Pass -DCMAKE_SUPPRESS_REGENERATION=ON in the cmake.yml template so every consumer of the template gets the fix automatically.

Symptom (intermittent, timing-dependent under load):
  error MSB3371: The file ARM64\\RelWithDebInfo\\ZERO_CHECK\\ZERO_CHECK.tlog\\unsuccessfulbuild
  cannot be created. The process cannot access the file ... because
  it is being used by another process.

Root cause: CMake's Visual Studio generator creates a ZERO_CHECK project that every other project in the solution depends on. When MSBuild builds projects in parallel (/m), multiple workers race to write the same tlog file.

CMAKE_SUPPRESS_REGENERATION=ON removes ZERO_CHECK entirely. The only trade-off is that cmake no longer auto-reruns when CMakeLists.txt changes, which is fine in CI (the build directory is freshly configured by this very step every job). No-op for Ninja/Makefile generators.

### DIFF
--- a/pipeline_templates/tasks/cmake.yml
+++ b/pipeline_templates/tasks/cmake.yml
@@ -80,7 +80,19 @@ steps:
         
         Push-Location $buildDir
         try {
-          $fullArgs = @($sourceDir) + ($cmakeArgs -split ' ') + @('-G', "$(cmakeGenerator)", '-A', "$(cmakeArch)")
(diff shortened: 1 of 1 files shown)


BertKleewein (approved):
:shipit:

PR #396, opened 2026-04-26
CETCOMAT is not ARM64 compatible

anporumb:


overlay_ports/pkgconf/portfile.cmake
```diff
@@ -6,8 +6,14 @@ vcpkg_from_github(
...
-set(OPTIONS "-Dc_args=/QSpectre /guard:cf /W3")
-list(APPEND OPTIONS "-Dc_link_args=/CETCOMPAT /guard:cf")
+if(VCPKG_TARGET_ARCHITECTURE STREQUAL "x64" OR VCPKG_TARGET_ARCHITECTURE STREQUAL "x86")
+    # /QSpectre and /CETCOMPAT are only supported on x86/x64; ARM64 rejects them.
+    set(OPTIONS "-Dc_args=/QSpectre /guard:cf /W3")
+    list(APPEND OPTIONS "-Dc_link_args=/CETCOMPAT /guard:cf")
+else()
+    set(OPTIONS "-Dc_args=/guard:cf /W3")
+    list(APPEND OPTIONS "-Dc_link_args=/guard:cf")
+endif()
```
dcristoloveanu, line 16: I would actually do this the other way around: if ARM64 then do something, else for anything else do the default thingies. /consider #Closed

dcristoloveanu (dismissed):
:shipit:

overlay_ports/pkgconf/portfile.cmake
anporumb: implemented.

dcristoloveanu (approved):
:shipit:

PR #402, opened 2026-05-06
[MrBot] Add MAX_STAGE param; make ARM64 a blocking, test-running leg

anporumb:
Replace the legacy `run_arm64_tests` / `is_optional` / `skip_tests` parameters in build_all_flavors.yml and build_and_run_tests.yml with a single MAX_STAGE knob (BUILD / UNIT_TESTS / ALL, default UNIT_TESTS) that matches the convention already in use by EBS / ELS / GEO. After consumers bump their deps/c-build-tools pin, ARM64 legs symmetrically run unit tests (CTest, VSTest, AppVerifier when configured) and are blocking.

Changes:

- build_all_flavors.yml
  - Add MAX_STAGE parameter (string enum, default UNIT_TESTS).
  - Delete run_arm64_tests parameter and all references.
  - Stop forcing is_optional=true and skip_tests=true on ARM64.
  - Pipe MAX_STAGE down to build_and_run_tests.yml.
  - AppVerifier exclusion: skip APP_VERIFIER=ON whenever MAX_STAGE=BUILD (no point building an AppVerifier flavor that won't run tests).

- build_and_run_tests.yml
  - Add MAX_STAGE parameter; remove is_optional and skip_tests.
  - Test-execution gating becomes a positive match on UNIT_TESTS / ALL.
  - Drop the job-level continueOnError and the per-task ARM64 continueOnError on the VSTest task.
  - Make /collect:"Code Coverage;Format=Xml" and the subsequent PublishCodeCoverageResults step conditional on ARCH_TYPE != ARM64. tasks/vstest.yml currently appends otherConsoleOptions verbatim and does not honor its own codeCoverageEnabled parameter, so without this guard ARM64 VSTest would request a coverage collector that is not validated on the ARM64 build pools. Pre-existing bug exposed by enabling ARM64 VSTest; full wrapper cleanup is a follow-up.

### DIFF
--- a/pipeline_templates/build_all_flavors.yml
+++ b/pipeline_templates/build_all_flavors.yml
@@ -11,7 +11,8 @@ parameters:
     type: object
     default: ["ON", "OFF"]
 
-  # ARCH_TYPE_VALUES - architectures to build. ARM64 builds are optional (non-blocking).
+  # ARCH_TYPE_VALUES - architectures to build. Both x64 and ARM64 are blocking
+  # by default; tests run on both unless the caller passes MAX_STAGE=BUILD.
   - name: ARCH_TYPE_VALUES
     type: object
     default: ["x64", "ARM64"]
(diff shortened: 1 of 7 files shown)


pipeline_templates/build_all_flavors.yml
```diff
@@ -66,11 +67,21 @@ parameters:
     type: string
     default: "Azure-MessagingStore-Win-ARM64-Dpldsv6-v01"
 
-  # Controls whether tests are executed on ARM64 builds. When false (default),
-  # ARM64 builds compile tests but do not run them.
-  - name: run_arm64_tests
-    type: boolean
-    default: false
+  # MAX_STAGE controls how far the pipeline runs. Matches the convention used
+  # by EBS / ELS / GEO so callers can use one knob across all repos.
```
mattdurak: maybe "EBS/ELS/GEO" mentions in c-build-tools should be removed... #Closed

mattdurak (dismissed):
:shipit:

dcristoloveanu (dismissed):
:shipit:

mattdurak (dismissed):
:shipit:

pipeline_templates/build_all_flavors.yml
anporumb: Replaced with generic "downstream".

mattdurak (approved):
:shipit:

PR #405, opened 2026-05-08
Support consumers with internal-ADO submodules and C# projects with private NuGet feeds

parth21999:
## Summary

Three independent template changes that unblock consumers with shapes c-build-tools hasn't seen before. **Default behavior is unchanged for all current consumers** — every change is no-op unless the consumer's repo shape benefits.

### 1. `submodules: true` on `checkout: self`

In both `build_setup_steps.yml` and `run_repo_validation.yml`: replace `submodules: false` + a separate `git submodule update --init --depth 1 --jobs 20` script step with `submodules: true` directly on the `checkout: self` step.

**Why**: the script-step pattern had no auth headers configured, which only works for `github.com` submodules (anonymous clone). Internal-ADO submodules at `msazure.visualstudio.com/...` URLs require per-URL `System.AccessToken` auth headers, which ADO's Checkout task injects automatically when `submodules: true`. The manual script step has no equivalent.

**No regression for github-only consumers**: same end state (submodules cloned shallowly via different mechanism). The injected auth headers are no-ops for `github.com` URLs since git just uses anonymous access there.

### 2. New `Stage repo-root nuget.config for msbuild restore` PowerShell step

In both `build_setup_steps.yml` and `run_repo_validation.yml`. Copies `$(Build.SourcesDirectory)/nuget.config` to `$(Build.BinariesDirectory)/c/` if it exists.

**Why**: msbuild's `/t:restore` walk-up search for `nuget.config` starts at the `.csproj` directory and walks upward. The solution lives at `$(Build.BinariesDirectory)/c/`, not under `$(Build.SourcesDirectory)`, so a `nuget.config` at the repo root is otherwise invisible to restore. Copying it under the binaries dir puts it on the walk-up path.

**No regression for consumers without a repo-root `nuget.config`**: the conditional copy just logs `No nuget.config; skipping` and returns. This is essential for consumers with C# projects that depend on private NuGet feeds declared in a repo-root `nuget.config`.

### 3. New `publish_artifacts_on_success: bool` parameter (default `false`)

In `build_and_run_tests.yml` and `build_all_flavors.yml`. When `true`, publishes `$(Build.BinariesDirectory)/<build_configuration>` as a pipeline artifact named `<build_configuration>` (e.g. `Debug`) on success.

**Why**: for consumers whose pipeline has separate test/validate jobs depending on the build artifact. The existing `build_teardown_steps.yml` only publishes the **whole** binaries dir on **failure** — useful for diagnostics, not for downstream consumption.

Default `false` → no extra step generated → zero impact for consumers without downstream artifact dependencies.

## Validation

End-to-end runs on both shapes of consumer:

| Repo | Build | Result | Proves |
|------|-------|--------|--------|
| zrpc PR #15664572 | [#163532935](https://msazure.visualstudio.com/One/_build/results?buildId=163532935) | **18/18 ✅** | github-only consumer without `nuget.config` and without downstream artifact consumer sees no behavioral change |
| Azure-Messaging-GeoReplication PR #15650373 | [#163505770](https://msazure.visualstudio.com/One/_build/results?buildId=163505770) | **38/40 ✅** | consumer with internal-ADO submodules + private `nuget.config` + downstream .NET test jobs gets unblocked. The 2 failures are pre-existing `.NET` test `TestCleanup` leak-check flakes unrelated to these changes — template-driven jobs (`build_and_run_tests`, `repo_validation`, `validate_nuget`) all green. |


pipeline_templates/build_setup_steps.yml
```diff
@@ -21,20 +21,51 @@ parameters:
...
-  # Shallow clone - only need the code at this commit, not full history
+  # Shallow clone - only need the code at this commit, not full history.
+  # submodules: true (instead of false + a separate 'git submodule update'
+  # script step) lets ADO's checkout task inject System.AccessToken-based
+  # auth headers for every submodule URL, not just the outer repo URL.
+  # This is required for any consumer that has internal-ADO submodules;
+  # the previous 'manual submodule update script' pattern had no auth
+  # headers at all and only worked for github.com submodules (anonymous
+  # clone). Existing github.com submodules clone fine either way since
+  # they're anonymous.
```
parth21999: Make this comment one line. just about the access token #Resolved

pipeline_templates/run_repo_validation.yml
```diff
@@ -41,16 +41,17 @@ jobs:
   steps:
   - template: get_machine_information.yml
 
-  # Shallow clone - only need the code at this commit, not full history
+  # Shallow clone - only need the code at this commit, not full history.
+  # submodules: true is required for any consumer that has internal-ADO
+  # submodules; the ADO Checkout task injects auth headers per-submodule-URL,
+  # while the 'submodules: false + manual git submodule update script'
+  # pattern only works for github.com (anonymous) submodules. github-only
+  # consumers see no behavioral change.
```
parth21999: Make comment short
 #Resolved

pipeline_templates/build_setup_steps.yml
```diff
@@ -21,20 +21,51 @@ parameters:
...
+  # Stage a repo-root nuget.config (if present) into the binaries dir so
+  # msbuild's integrated /t:restore discovers it. msbuild's walk-up search
+  # for nuget.config starts at the .csproj directory and walks upward;
+  # the SOLUTION lives at $(Build.BinariesDirectory)/c, not under the
+  # sources dir, so a nuget.config at the repo root is otherwise
+  # invisible to restore. Copying it under the binaries dir puts it on
+  # the walk-up path. No-op for consumers without a repo-root
+  # nuget.config (the conditional copy just skips). Case-preserving
+  # using Get-ChildItem -Filter so a config named NuGet.config or
+  # nuget.config keeps its original case.
```
parth21999: Make comment shorter
 #Resolved

pipeline_templates/run_repo_validation.yml
```diff
@@ -41,16 +41,12 @@ jobs:
   steps:
   - template: get_machine_information.yml
 
-  # Shallow clone - only need the code at this commit, not full history
+  # submodules: true puts the System.AccessToken on the auth header for every submodule URL (not just the outer repo URL), so consumers with internal-ADO submodules work.
```
anporumb: >true [](http://example.com/codeflow?start=16&length=4)

So how did ZRPC build until now? ZRPC is a repo that uses METRICS (another  ADO repo).

So the question here is what is it we are trying to solve with the exact change?

[for users outside Microsoft - ZRPC/METRICS etc - internal names to some non-public repoes) #Closed

pipeline_templates/run_repo_validation.yml
```diff
@@ -87,6 +83,26 @@ jobs:
   - ${{ if parameters.setup_nuget }}:
     - template: setup_nuget_tools.yml
 
+  # Stage a repo-root nuget.config (if present) into the binaries dir so
+  # msbuild's integrated /t:restore discovers it. See build_setup_steps.yml
+  # for rationale. No-op if no repo-root nuget.config exists.
```
anporumb: >No-o [](http://example.com/codeflow?start=19&length=4)

The way of addressing "let's have nuget.config placed in build binaries directory" is wrong.

CMake needs it, so CMake should copy if (if it exists).

Then you don't need to change anything here, you don't need to create a folder in which AFTER CMake would so something etc.

Have CMake copy the file, just like any other needed build artifact. #Resolved

pipeline_templates/build_setup_steps.yml
```diff
@@ -21,20 +21,35 @@ parameters:
...
+        $binariesC = "$(Build.BinariesDirectory)/c"
+        $candidate = Get-ChildItem -Path $sourcesDir -Filter 'nuget.config' -File -ErrorAction SilentlyContinue | Select-Object -First 1
+        if (-not $candidate) {
+          Write-Host "No nuget.config at $sourcesDir; skipping."
+          return
+        }
+        New-Item -ItemType Directory -Force -Path $binariesC | Out-Null
+        $dst = Join-Path $binariesC $candidate.Name
+        Copy-Item -LiteralPath $candidate.FullName -Destination $dst -Force
+        Write-Host "Staged $($candidate.FullName) -> $dst"
```
anporumb: I think I've commented about this, in run_repo_validation, sounds very "copy&paste".

And let's remember: copy and paste is a design mistake. #Closed

pipeline_templates/build_and_run_tests.yml
```diff
@@ -62,6 +62,16 @@ parameters:
   - name: setup_nuget
     type: boolean
     default: false
+  # When true, publish $(Build.BinariesDirectory)/<build_configuration> as
+  # a pipeline artifact named <build_configuration> on success (in addition
+  # to the failure-time whole-binaries publish handled by build_teardown_steps.yml).
```
anporumb: look, when I see "in addition" looks like we're going to publish some things, under some conditions "twice".

Can we please, please, PLEASE, for the love of Australia East, which publishes in US, do not do that? #Closed

pipeline_templates/build_and_run_tests.yml
```diff
@@ -241,6 +251,21 @@ jobs:
...
 
+  # Optional: publish the configuration's build output as a named pipeline
+  # artifact on success, so downstream jobs in the same pipeline can
+  # consume it via DownloadPipelineArtifact. The artifact is keyed by
+  # build_configuration (e.g. 'Debug', 'RelWithDebInfo') so downstream
+  # consumers can use a stable, simple name. Default-off keeps current
+  # behavior for consumers that don't have downstream artifact dependencies.
+  - ${{ if parameters.publish_artifacts_on_success }}:
+    - task: PublishPipelineArtifact@1
+      displayName: 'Publish ${{ parameters.build_configuration }} build artifact'
```
anporumb: >${{ parameters.build_configuration }} [](http://example.com/codeflow?start=28&length=37)

Maybe this will work, maybe not. I would suggest looking at "SUFFIX" whcih (bad name) wants to create a unique string for these configurations.

What happens if you build ARM64 and x64 for Debug? Do you upload twice, some config name? #Closed

anporumb: the "maybe not" really needs to be tested across all repoes so we don't get unexpected surprises when propagating and then we have to do ... retakes. Bot will help with that.

pipeline_templates/build_and_run_tests.yml
anporumb: I've looked briefly at build_teardown_steps and it looks like it publishes anyway. There's no conditional there. So why publish twice ro is there something I do not understand here?

pipeline_templates/build_all_flavors.yml
```diff
@@ -88,6 +88,14 @@ parameters:
     type: boolean
     default: false
 
+  # Forwarded to build_and_run_tests.yml. When true, publish the
+  # configuration's binaries dir as a named pipeline artifact so
+  # downstream jobs in the same pipeline can consume it. See
+  # build_and_run_tests.yml for details.
+  - name: publish_artifacts_on_success
+    type: boolean
+    default: false
```
anporumb, line 110: Actually very unclear why this is needed.

I will be waiting for author here because I need to see where this is going. #Closed

anporumb (changes requested):
:clock1:

pipeline_templates/run_repo_validation.yml
parth21999: zrpc does not use metrics.

Here is ZRPC's .gitmodules:

```
[submodule "deps/umock-c"]
 path = deps/umock-c
 url = [https://github.com/Azure/umock-c](https://github.com/Azure/umock-c)
[submodule "deps/clds"]
 path = deps/clds
 url = [https://github.com/Azure/clds.git](https://github.com/Azure/clds.git)
[submodule "deps/mimalloc"]
 path = deps/mimalloc
 url = [https://github.com/microsoft/mimalloc](https://github.com/microsoft/mimalloc)
[submodule "deps/c-build-tools"]
 path = deps/c-build-tools
 url = [https://github.com/Azure/c-build-tools](https://github.com/Azure/c-build-tools)
[submodule "deps/macro-utils-c"]
 path = deps/macro-utils-c
 url = [https://github.com/Azure/macro-utils-c](https://github.com/Azure/macro-utils-c)
[submodule "deps/c-logging"]
 path = deps/c-logging
 url = [https://github.com/Azure/c-logging](https://github.com/Azure/c-logging)
[submodule "deps/ctest"]
 path = deps/ctest
 url = [https://github.com/Azure/ctest](https://github.com/Azure/ctest)
[submodule "deps/c-testrunnerswitcher"]
 path = deps/c-testrunnerswitcher
 url = [https://github.com/Azure/c-testrunnerswitcher](https://github.com/Azure/c-testrunnerswitcher)
[submodule "deps/c-pal"]
 path = deps/c-pal
 url = [https://github.com/Azure/c-pal](https://github.com/Azure/c-pal)
[submodule "deps/c-util"]
 path = deps/c-util
 url = [https://github.com/Azure/c-util](https://github.com/Azure/c-util)
[submodule "deps/smhasher"]
 path = deps/smhasher
 url = [https://github.com/hcoona/smhasher.git](https://github.com/hcoona/smhasher.git)
[submodule "deps/vcpkg"]
 path = deps/vcpkg
 url = [https://github.com/microsoft/vcpkg.git](https://github.com/microsoft/vcpkg.git)
[submodule "deps/libuv"]
 path = deps/libuv
 url = [https://github.com/libuv/libuv.git](https://github.com/libuv/libuv.git)
[submodule "deps/azure-messaging-ai-context"]
 path = deps/azure-messaging-ai-context
 url = [https://github.com/Azure/azure-messaging-ai-context.git](https://github.com/Azure/azure-messaging-ai-context.git)
```

pipeline_templates/build_and_run_tests.yml
parth21999: [https://github.com/Azure/c-build-tools/blob/master/pipeline_templates/build_teardown_steps.yml](https://github.com/Azure/c-build-tools/blob/master/pipeline_templates/build_teardown_steps.yml)

```
# Reusable step template for build teardown: stop trace collection, publish artifacts on failure, and clean up.
# Used by build_and_run_tests.yml for both the main test job and the AppVerifier job.

parameters:
  - name: repo_root_override
    default: $(Build.SourcesDirectory)/deps/c-build-tools
  - name: ARCH_TYPE
    default: "x64"
  - name: build_configuration
    default: "Debug"
  - name: artifact_name_suffix
    type: string

steps:
  - template : stop_logman.yml
    parameters:
      ${{ if ne(parameters.repo_root_override, '') }}:
        repo_root: ${{ parameters.repo_root_override }}

  - task: PublishPipelineArtifact@1
    displayName: 'Publish ${{ parameters.ARCH_TYPE }} ${{ parameters.build_configuration}} Artifacts'
    inputs:
      targetPath: $(Build.BinariesDirectory)
      artifactName: ${{ parameters.artifact_name_suffix }}_artifacts
      parallel: true
    condition: failed()    !!!!

  - template : clean_ado_folders.yml
```

parth21999: [bot response]

`build_teardown_steps.yml` line 26 has `condition: failed()` on its `PublishPipelineArtifact@1`. The new step in `build_and_run_tests.yml` has `condition: succeeded()`. A job is either success or failure, never both, so the two publishes are **mutually exclusive** — exactly one runs per job, and it's `failed()` exactly when something went wrong (which is when we want the whole binaries dir for diagnostics) and `succeeded()` exactly when we want to hand the per-config output to downstream jobs.

Reworded both comments in `bb9df83` to spell that out instead of saying "in addition", which was what made it read like a double-publish.

pipeline_templates/run_repo_validation.yml
parth21999: Removed.

pipeline_templates/build_and_run_tests.yml
parth21999: Updated comment for clarity.

pipeline_templates/build_and_run_tests.yml
parth21999: Thanks for the suggestion, done.

pipeline_templates/build_all_flavors.yml
parth21999: Updated comment in build_and_run_tests.yml

TLDR: GEO needs the build artifacts published on success because the .NET test pipelines consume them.

pipeline_templates/build_setup_steps.yml
parth21999: Refactored into common_setup_steps.yml

pipeline_templates/build_and_run_tests.yml
```diff
@@ -62,6 +62,20 @@ parameters:
   - name: setup_nuget
     type: boolean
     default: false
+  # When true, publish $(Build.BinariesDirectory)/<build_configuration> as
+  # a pipeline artifact named <BUILD_SUFFIX>_build_output on success.
```
anporumb: ><BUILD_SUFFIX>_build_output [](http://example.com/codeflow?start=30&length=27)

"it is in the name".

If it called "suffix" - why does it appear at the beginning? That which appears at the beginning is called PREFIX. #Closed

pipeline_templates/build_and_run_tests.yml
```diff
@@ -62,6 +62,20 @@ parameters:
...
+  # The artifact name is keyed on BUILD_SUFFIX (not just build_configuration)
+  # so flavors that share a build_configuration -- e.g. x64 + ARM64 Debug --
+  # don't collide on the artifact name. Mutually exclusive with
+  # build_teardown_steps.yml's failure-time publish (`condition: failed()`)
+  # -- never both, since a job is either success or failure. For consumers
+  # whose downstream jobs need the build output (e.g. .NET test jobs that
+  # DownloadPipelineArtifact the build output). Default false -> step is
+  # skipped -> no cost for consumers without downstream artifact
```
anporumb: bot tends to be verbose with its comments, I had to instruct mine to abstain from self-aggrandizing. #Closed

anporumb (dismissed):
:shipit:

pipeline_templates/build_and_run_tests.yml
parth21999: Fixed.

pipeline_templates/build_and_run_tests.yml
parth21999: Fixed.

anporumb (approved):
:shipit:

PR #407, opened 2026-05-14
Configure Azure DevOps CLI defaults during setup

parth21999:
After authentication, auto-configure az devops defaults for organization and project. Without this, users must manually run the configure command before the propagation script works.


anporumb (approved):
:shipit:

PR #412, opened 2026-05-23
Fix MSB3836 binding redirect warning for System.Collections.Immutable

dcristoloveanu:
## Summary

Update the explicit binding redirect for System.Collections.Immutable in sarif_results_checker/App.config from version 1.2.3.0 to 9.0.0.7 to match the auto-generated binding redirect.

## Problem

The App.config had an explicit binding redirect to version 1.2.3.0. MSBuild now auto-generates a redirect to version 9.0.0.7 (from system.collections.immutable 9.0.7 package), causing MSB3836 warning:

The explicit binding redirect on System.Collections.Immutable conflicts with an autogenerated binding redirect.

## Fix

Updated explicit redirect oldVersion range to 0.0.0.0-9.0.0.7 and newVersion to 9.0.0.7 to match what MSBuild auto-generates.

Fixes: https://msazure.visualstudio.com/One/_workitems/edit/38082643


anporumb (approved):
:shipit:

PR #417, opened 2026-06-09
Add Azure VM, IMDS network and NIC details to Machine Information

anporumb:
Enrich the Machine Information diagnostic step with Azure Instance Metadata Service data: VM name, Azure SKU (vmSize), region/zone/fault domain, provisioned image version (exactVersion), ephemeral OS, VMSS name and resource ids, plus per-NIC IMDS addressing and in-box network adapter capabilities including Accelerated Networking detection.

All IMDS access is gated behind a 1.5s TCP probe and uses proxy-disabled, hard-timeout HTTP calls, so the step never hangs or fails the pipeline on non-Azure or proxied agents. Existing continueOnError, condition always() and exit 0 guards are preserved.

### DIFF
--- a/pipeline_templates/get_machine_information.yml
+++ b/pipeline_templates/get_machine_information.yml
@@ -220,6 +220,140 @@ steps:
         }
         Write-Host ""
 
+        # --- Azure VM Metadata (IMDS) ---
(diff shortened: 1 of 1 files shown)


BertKleewein (approved):
:shipit:

PR #419, opened 2026-06-12
Add pool_demands param to build templates

parth21999:
## What

Adds one optional, default-noop parameter to the build templates so consumers can extend the build job's agent demands:

### `pool_demands` (object, default `[]`)
Extra agent demands appended to the build job pool's existing `msbuild` + `visualstudio` demands. Lets a consumer require additional agent capabilities — e.g. a local Service Fabric cluster for SF integration tests — without forking the template.

- `build_and_run_tests.yml`: appends `${{ each demand in parameters.pool_demands }}` to the pool `demands` list.
- `build_all_flavors.yml`: forwards `pool_demands` to the per-flavor `build_and_run_tests.yml` invocation.

Empty (default) = no extra demands appended → existing consumers (zrpc, c-pal, …) unaffected.

## Note
An earlier revision of this PR also added a `nuget_config_path` parameter for out-of-source NuGet restore. That was dropped: consumers already solve out-of-source restore by staging `nuget.config` into the CMake binary dir (`configure_file` → `${CMAKE_BINARY_DIR}`), which msbuild's restore walk-up then finds. No template change is needed for that, so only `pool_demands` remains.

## Validation
`pool_demands` was exercised end-to-end on an Azure-Messaging-GeoReplication test gate run (def 226370): the build-job pool block rendered the `runsOnLocal5MachineSFCluster` demand and all build/test jobs ran on matching agents.

### DIFF
--- a/pipeline_templates/build_all_flavors.yml
+++ b/pipeline_templates/build_all_flavors.yml
@@ -121,6 +121,21 @@ parameters:
     type: number
     default: 1200
 
+  # Forwarded to build_setup_steps.yml / run_repo_validation.yml. When set to a
+  # non-empty string, a dedicated DotNetCoreCLI@2 restore with this nuget.config
+  # runs before msbuild (which then runs /t:build only). Default empty keeps the
+  # combined /t:restore /t:build behavior. See build_setup_steps.yml for rationale.
+  - name: nuget_config_path
+    type: string
+    default: ''
+
+  # Forwarded to build_and_run_tests.yml. Extra pool demands (agent capabilities)
+  # required on top of the default msbuild + visualstudio for the per-flavor
+  # build+test jobs (e.g. a local Service Fabric cluster).
+  - name: pool_demands
+    type: object
+    default: []
+
 jobs:
 # Run repo validation and traceability once (not per-flavor)
 - ${{ if not(parameters.skip_repo_validation) }}:
@@ -129,6 +144,7 @@ jobs:
       additional_cmake_options: ${{ parameters.additional_cmake_options }}
       pool_name: ${{ parameters.pool_name_x64 }}
       setup_nuget: ${{ parameters.setup_nuget }}
+      nuget_config_path: ${{ parameters.nuget_config_path }}
 
 - ${{ each GBALLOC_LL_TYPE in parameters.GBALLOC_LL_TYPE_VALUES }}:
   - ${{ each FSANITIZE_TYPE in parameters.FSANITIZE_TYPE_VALUES }}:
@@ -160,6 +176,8 @@ jobs:
                 enable_hang_watchdog: ${{ parameters.enable_hang_watchdog }}
                 hang_watchdog_name_pattern: ${{ parameters.hang_watchdog_name_pattern }}
                 hang_dump_threshold_sec: ${{ parameters.hang_dump_threshold_sec }}
+                nuget_config_path: ${{ parameters.nuget_config_path }}
+                pool_demands: ${{ parameters.pool_demands }}
                 # Per-architecture pool selection AND effective MAX_STAGE.
                 # ARM64 -> arm64_max_stage (default UNIT_TESTS; caller-settable so
                 # repos migrate ARM64 test depth one at a time). x64 -> ALL (run
--- a/pipeline_templates/build_all_flavors.yml
+++ b/pipeline_templates/build_all_flavors.yml
@@ -121,17 +121,12 @@ parameters:
     type: number
     default: 1200
 
-  # Forwarded to build_setup_steps.yml / run_repo_validation.yml. When set to a
-  # non-empty string, a dedicated DotNetCoreCLI@2 restore with this nuget.config
-  # runs before msbuild (which then runs /t:build only). Default empty keeps the
-  # combined /t:restore /t:build behavior. See build_setup_steps.yml for rationale.
+  # Forwarded to build_setup_steps.yml / run_repo_validation.yml. See build_setup_steps.yml.
   - name: nuget_config_path
     type: string
     default: ''
 
-  # Forwarded to build_and_run_tests.yml. Extra pool demands (agent capabilities)
-  # required on top of the default msbuild + visualstudio for the per-flavor
-  # build+test jobs (e.g. a local Service Fabric cluster).
+  # Forwarded to build_and_run_tests.yml. Extra agent demands for the build+test jobs.
   - name: pool_demands
     type: object
     default: []
--- a/pipeline_templates/build_all_flavors.yml
+++ b/pipeline_templates/build_all_flavors.yml
@@ -121,11 +121,6 @@ parameters:
     type: number
     default: 1200
 
-  # Forwarded to build_setup_steps.yml / run_repo_validation.yml. See build_setup_steps.yml.
-  - name: nuget_config_path
-    type: string
-    default: ''
-
   # Forwarded to build_and_run_tests.yml. Extra agent demands for the build+test jobs.
   - name: pool_demands
     type: object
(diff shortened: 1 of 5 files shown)


pipeline_templates/build_all_flavors.yml
```diff
@@ -121,6 +121,21 @@ parameters:
     type: number
     default: 1200
 
+  # Forwarded to build_setup_steps.yml / run_repo_validation.yml. When set to a
+  # non-empty string, a dedicated DotNetCoreCLI@2 restore with this nuget.config
+  # runs before msbuild (which then runs /t:build only). Default empty keeps the
+  # combined /t:restore /t:build behavior. See build_setup_steps.yml for rationale.
```
parth21999: Make this comment shorter.

pipeline_templates/build_and_run_tests.yml
```diff
@@ -83,6 +83,19 @@ parameters:
...
+  # When set to a non-empty string, run a dedicated DotNetCoreCLI@2 restore using
+  # this nuget.config path before msbuild (which then runs /t:build only). Default
+  # empty -> combined /t:restore /t:build. See build_setup_steps.yml for rationale.
+  - name: nuget_config_path
+    type: string
+    default: ''
+  # Extra pool demands (agent capabilities) required on top of the default
+  # msbuild + visualstudio. Use for consumers whose tests need specific agent
+  # capabilities (e.g. a local Service Fabric cluster). The job queues until an
+  # agent in the pool advertising all listed capabilities is available.
```
parth21999: Make this comment shorter #Closed

pipeline_templates/build_and_run_tests.yml
```diff
@@ -83,6 +83,19 @@ parameters:
   - name: hang_dump_threshold_sec
     type: number
     default: 1200
+  # When set to a non-empty string, run a dedicated DotNetCoreCLI@2 restore using
+  # this nuget.config path before msbuild (which then runs /t:build only). Default
+  # empty -> combined /t:restore /t:build. See build_setup_steps.yml for rationale.
```
parth21999: shorter #Closed

pipeline_templates/build_setup_steps.yml
```diff
@@ -17,6 +17,15 @@ parameters:
     type: string
   - name: use_vld_option
     type: string
+  # When set to a non-empty string, run a dedicated DotNetCoreCLI@2 restore
+  # task using this nuget.config path before the msbuild build step (and run
+  # msbuild with /t:build only). Default: empty -> combined /t:restore /t:build
+  # (existing behavior). Set this when the consumer's internal-feed nuget.config
+  # cannot be found by NuGet's walk-up search from the generated solution under
+  # ${CMAKE_BINARY_DIR} (which is outside the source tree).
```
parth21999: shorter

pipeline_templates/build_setup_steps.yml
```diff
@@ -38,13 +47,31 @@ steps:
       buildDirectory: '$(Build.BinariesDirectory)/c'
       cmakeArgs: '${{ parameters.cmake_options }} ${{ parameters.use_vld_option }} -DGBALLOC_LL_TYPE:string=${{ parameters.GBALLOC_LL_TYPE }} -DCMAKE_BUILD_TYPE=${{ parameters.build_configuration }}'
 
-  # Use wrapper template for MSBuild - handles ARM64 native tools transparently
+  # Optional dedicated NuGet restore using an explicit nuget.config path. Needed
+  # when the generated solution lives under ${CMAKE_BINARY_DIR} and the consumer's
+  # nuget.config sits in the source tree -- msbuild's /t:restore walk-up never
+  # crosses that boundary.
```
parth21999: make this comment shorter #Closed

pipeline_templates/run_repo_validation.yml
```diff
@@ -25,6 +25,15 @@ parameters:
     type: boolean
     default: false
 
+  # When set to a non-empty string, run a dedicated DotNetCoreCLI@2 restore
+  # task using this nuget.config path before the msbuild build step (and run
+  # msbuild with /t:build only). Default: empty -> combined /t:restore /t:build.
+  # Set this when the consumer's internal-feed nuget.config cannot be found by
+  # NuGet's walk-up search from the generated solution under ${CMAKE_BINARY_DIR}.
```
parth21999: shorter

pipeline_templates/build_all_flavors.yml
parth21999: Shortened to 1-2 lines in f0c1274.

pipeline_templates/build_and_run_tests.yml
parth21999: Shortened to 1-2 lines in f0c1274.

pipeline_templates/build_and_run_tests.yml
parth21999: Shortened to 1-2 lines in f0c1274.

pipeline_templates/build_setup_steps.yml
parth21999: Shortened to 1-2 lines in f0c1274.

pipeline_templates/build_setup_steps.yml
parth21999: Shortened to 1-2 lines in f0c1274.

pipeline_templates/run_repo_validation.yml
parth21999: Shortened to 1-2 lines in f0c1274.

anporumb:
I don't think this is needed, is it
I need to understand WHY this is needed, when I can build everything out of source tree and so does EBS and ELS and GEO.

parth21999:
Removed. Still need the pool_demands

---
In reply to: [4696782227](https://github.com/Azure/c-build-tools/pull/419#issuecomment-4696782227) [](http://example.com/codeflow?ancestors=4696782227)

pipeline_templates/build_and_run_tests.yml
```diff
@@ -99,6 +112,8 @@ jobs:
     demands:
     - msbuild
     - visualstudio
+    - ${{ each demand in parameters.pool_demands }}:
+      - ${{ demand }}
```
anporumb, line 108: ok, so this makes the demands of the gate to be configurable. All nice, but what purpose does it serve?

We have a very limited number of pools. Pools have 1 (one) type of machine with 1(one) config.

So whatever the agents will advertise as "capabilities" to be matches with the "demands" is never changing.

At this point either a job runs or it doesn't (ever - on any of the pools machines).

So why do we need this?

anporumb (changes requested):
:clock1:

pipeline_templates/build_and_run_tests.yml
parth21999: build_geo.yml (now removed) specified pool demands.

That file was removed when I changed the geo gate to use the c-build-tools templates. Since the templates did not support demands, that capability was lost. 

With this PR, I was trying to restore the ability of the geo gate to specify pool demands.

Agent capabilities can change if the config is modified. Pool demands will cause the gate to fail with a nice error if the gate tries to queue a job on an agent that does not meet the demands. Might make debugging easier.

anporumb: Yes, but the question was "why do we need this". Note: present tense.

Was not "why we might need this" or "why we will need this"

There is nothing in our pipelines that has "different" demands or because we do not exposes differente capabilities. Our agents/images/pools are tiny and the image always has "everything". If you need some capability - go and change the image definition. It is so much easier to have 1 definition AllInOneWonder than 10 images each with its own capabilities. We cannot afford the effort to maintain multiple images.

So I still don't understand why the need. The only gain would be the "make debugging easier" - very little ROI predicted.

When the need will ne stringent and well articulated/reasonable/unavoidable we can re-evaluate this.

Anyway - this is my last comment here, I am not adding any value anymore.

PR #420, opened 2026-06-15
shift pool from Azure-MessagingStore-Win-ARM64-Dpldsv6-v01 to Azure-M…

anporumb:
…essagingStore-Win-ARM64-Dpldsv6-v04


jebrando (approved):
:shipit:

PR #431, opened 2026-07-02
[MrBot][build] Retry git checkout steps on transient network failures

mattdurak:
## Summary
Broadly applies the transient git-checkout retry fix from [ELS PR 16293811](https://msazure.visualstudio.com/One/_git/Azure-Messaging-ElasticLog/pullrequest/16293811) to this repo's pipeline YAML.

Every pipeline `checkout:` step now sets `retryCountOnTaskFailure: 3`, so a transient network failure during Get Sources / submodule fetch (e.g. `curl 56 Recv failure: Connection was reset` -> `fatal: early EOF`) is retried automatically instead of failing the leg.

These are the shared `pipeline_templates/*` consumed by the C-library repos (c-pal, c-util, clds, zrpc, etc.), so this covers checkout retry for all consumers once they pick up the new `c_build_tools` ref.

## Change (build-YAML only)
- Added `retryCountOnTaskFailure: 3` to every git `checkout:` step, each preceded by a brief, consistent comment.

## Verification
- YAML parse: PASS on all changed files.
- Purely additive (2 lines per checkout step, 0 deletions).

### DIFF
--- a/pipeline_templates/build_linux.yml
+++ b/pipeline_templates/build_linux.yml
@@ -24,6 +24,8 @@ jobs:
   steps:
   # Shallow clone - only need the code at this commit, not full history
   - checkout: self
+    # Retry on failure to handle transient network connection failures (curl 56 / early EOF).
+    retryCountOnTaskFailure: 3
     submodules: false
     clean: false
     fetchDepth: 1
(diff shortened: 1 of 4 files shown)


anporumb (approved):
:shipit:

PR #440, opened 2026-07-27
Require two human reviewers on bot-authored pull requests [Ready]

mattdurak:
A pull request opened by our build agent is missing the human who would normally have written and vetted the change, so a single approval would be the only human judgement ever applied to it. This adds a workflow that requires **two distinct human approvals** on bot-authored PRs and passes immediately on everyone else's, leaving normal review untouched.

### Why a workflow instead of a branch rule

GitHub can't condition an approval count on the PR author, and a required check that never reports leaves a PR stuck on *"Expected — waiting for status to be reported"* forever. So it can't be posted only on bot PRs — it has to run on every PR and decide for itself.

### How it decides

- Passes immediately unless the PR author is a **bot**. It keys on the author's `Bot` type rather than a list of named bots, so renaming an app can't silently reopen the gap — getting it wrong can only ever be too strict. Exempting a specific bot is an explicit opt-in (`exempt-bots`).
- Counts each reviewer's most recent *decisive* review (`APPROVED` / `CHANGES_REQUESTED`), grouped by reviewer: one person approving twice is still one human, a later `CHANGES_REQUESTED` revokes, a later `COMMENTED` does not, and bot approvals don't count.
- Needs only `pull-requests: read`, and **never checks out the PR's code** — so running on `pull_request_review`, which carries a writable token, can't execute anything from the branch under review.

### Shared definition

This file is the definition for all the repos the agent touches. It declares `workflow_call` so other repositories can use it from a ~12-line caller, and `pull_request` so it also gates this repository's own PRs. Inputs: `required-approvals` (default 2) and `exempt-bots` (default none).

Verified live on both paths: this PR (self-trigger) and `Azure/umock-c#514` (caller).

### Before requiring the check

Merge this first, **then** add the status check — requiring it while the file is absent from `master` would block every open PR in the repo on a check that can't run. The context name differs by path, and branch protection matches exactly:

| Repo | Context to require |
| --- | --- |
| this one (runs directly) | `two-human-reviewers` |
| caller repos | `two-human-reviewers / two-human-reviewers` |


.github/workflows/two-human-reviewers.yml
```diff
@@ -0,0 +1,100 @@
...
+        env:
+          # github.token rather than secrets.GITHUB_TOKEN so this works without the caller passing
+          # `secrets: inherit`.
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR: ${{ github.event.pull_request.number }}
+          AUTHOR: ${{ github.event.pull_request.user.login }}
+          AUTHOR_TYPE: ${{ github.event.pull_request.user.type }}
+          # `inputs` is empty when this runs on its own repository's PRs rather than being called.
+          EXEMPT_BOT_LOGINS: ${{ inputs.exempt-bot-logins || '' }}
```
mattdurak: >EXEMPT_BOT_LOGINS [](http://example.com/codeflow?start=10&length=17)

why do we call this "logins"? Do a pass over this file to make sure it's still consistent #Closed

mattdurak: 🤖 MrBot: Renamed it to `exempt-bots`. You're right that "logins" was naming the *format* of the values rather than the thing being configured, and it was the only spot in the file using that vocabulary — everywhere else says "bot" / "bot-authored". The input description now carries the format detail instead ("GitHub account names ... e.g. `dependabot[bot]`").

The consistency pass turned up one real drift beyond the naming: several comments still asserted **two** approvals as a fact, but that became configurable when I added the `required-approvals` input during the reusable-workflow restructure. Those now point at the input rather than restating a number that a caller can override. Also unified "a list of bot logins" → "a list of named bots" and renamed the loop variable.

Re-validated after the rename: the reusable path (this PR, run 30369831325) and the caller path (`Azure/umock-c#514`) are both green, and the seven author-gating cases still behave — human passes, both the current and old bot names gate, exempt matches with or without surrounding spaces.

See commit 6e54483.

jebrando (commented):
Two issues from reviewing the gate logic — flagged inline. Both concern the *reusability* goal and the trust model of the approval count; details in the comments.

.github/workflows/two-human-reviewers.yml
```diff
@@ -0,0 +1,102 @@
...
+      exempt-bots:
+        description: >-
+          Comma-separated GitHub account names of bots that may merge under the repository's normal
+          rules, e.g. "dependabot[bot]".
+        type: string
+        required: false
+        default: ''
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+  pull_request_review:
```
jebrando: **Reusable callers won't re-trigger on approvals → the required check can stick failing.**

When this workflow is *called* via `workflow_call`, its own `pull_request` / `pull_request_review` triggers are ignored — only the **caller's** events drive execution. So the `pull_request_review: [submitted, dismissed]` trigger here only applies to c-build-tools' own PRs.

A consuming repo whose thin caller workflow declares just `on: pull_request` will run the gate once at open (0 approvals → fail) and then **never re-run when humans approve**, leaving the required check stuck failing ("Expected — waiting for status to be reported").

The reusable workflow can't enforce its callers' triggers, so it'd help to ship a caller template and document that callers **must** include both:

```yaml
on:
  pull_request:
    types: [opened, reopened, synchronize, ready_for_review]
  pull_request_review:
    types: [submitted, dismissed]
``` #Resolved

File: .github/workflows/two-human-reviewers.yml [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+              fi
+            done
+          fi
+
+          # A reviewer's standing verdict is their most recent APPROVED / CHANGES_REQUESTED review;
+          # COMMENTED and PENDING reviews say nothing either way and DISMISSED ones no longer count.
+          # Group by reviewer so re-approving the same PR twice can never count as two people.
+          approvals=$(gh api --paginate "repos/$REPO/pulls/$PR/reviews" | jq -s '
+            add // []
+            | map(select(.user.type != "Bot"))
```
jebrando: **No reviewer-trust filter — drive-by / sock-puppet approvals count.**

This keeps any `.user.type != "Bot"` APPROVED review but never checks `author_association` or repo permission. On a public repo (this one), *any* GitHub user can submit an approving review — so two throwaway accounts produce two distinct human `user.id`s and the gate passes. GitHub's own required-approval counting deliberately ignores approvals from users **without** write access; this gate doesn't.

Consider restricting to trusted associations, mirroring how GitHub itself counts:

```jq
| map(select(.author_association == "OWNER" or .author_association == "MEMBER" or .author_association == "COLLABORATOR"))
```

(End-to-end exploitability also depends on native branch protection, but the gate shouldn't count untrusted approvals as "humans.") #Resolved

jebrando (dismissed):
:shipit:

.github/workflows/two-human-reviewers.yml
mattdurak: 🤖 MrBot: Fixed in 2ac152c — this was exploitable, not theoretical: c-build-tools is **public**, so any GitHub account can submit an approving review, and two throwaway accounts gave two distinct `user.id`s and satisfied the gate. That is precisely the outcome it exists to prevent.

Approvals are now restricted to `OWNER` / `MEMBER` / `COLLABORATOR`, matching how GitHub counts its own required approvals. `CONTRIBUTOR` and `FIRST_TIME_CONTRIBUTOR` deliberately don't count — they carry no write access.

The filter runs before the grouping so an untrusted reviewer can't occupy a group at all.

Verified against 11 payload scenarios, including `two sock puppets (NONE)` → **0**, `CONTRIBUTOR does not count` → 1, and `MEMBER + outsider` → 1, alongside the existing stale-commit and dismissal cases.

.github/workflows/two-human-reviewers.yml
mattdurak: 🤖 MrBot: Correct on the mechanism, and documented in 2ac152c.

The shipped caller template (`two-human-reviewers.caller.yml`, alongside the reusable file) already declares both triggers, and the rollout script writes every repo's caller from it — so the repos we onboard get this right by construction. `Azure/umock-c#514` is the live example.

The gap was that nothing *said* so, which is a trap for anyone hand-writing a caller: the symptom is a required check stuck failing with no obvious cause. Both files now state it — the reusable one notes that a called workflow's own triggers are ignored and that callers must declare `pull_request_review`, and the caller template carries a "keep BOTH triggers" note at the point someone would be tempted to trim it.

One clarification on the symptom: a caller missing `pull_request_review` yields a check that *reports and fails* rather than one stuck on "Expected — waiting for status", since the run does happen at open. Same end result — unmergeable until an unrelated push re-triggers it — but it fails visibly rather than hanging.

.github/workflows/two-human-reviewers.yml
```diff
@@ -0,0 +1,143 @@
...
+          # with a GitHub account can submit an approving review, and counting those would let two
+          # throwaway accounts satisfy the gate. This mirrors how GitHub counts its own required
+          # approvals, which ignores reviewers without write access.
+          #
+          # An approval counts only for the commit it was given on: the bot pushes to its own PR, so an
+          # approval of an earlier commit must not carry over to code no human has looked at.
+          approvals=$(gh api --paginate "repos/$REPO/pulls/$PR/reviews" | jq -s --arg head "$head_sha" '
+            add // []
+            | map(select(.user.type != "Bot"))
+            | map(select(.author_association == "OWNER"
```
darobs: MrBot says this treats  MEMBER / COLLABORATOR  as proof of write access. Org members or read-only collaborators could satisfy the gate. It should query effective permission and require  write ,  maintain , or  admin.
As a project, I don't think we have any read-only member/collaborator access, so this may not be an issue, but it's worth noting.

darobs (dismissed):

.github/workflows/two-human-reviewers.yml
```diff
@@ -0,0 +1,143 @@
...
+                         or .author_association == "MEMBER"
+                         or .author_association == "COLLABORATOR"))
+            | map(select(.state == "APPROVED" or .state == "CHANGES_REQUESTED" or .state == "DISMISSED"))
```
jebrando: Bot - reviewer trust check does not verify write access - Read-only organization members or collaborators can still count. Check actual permission or GraphQL’s  authorCanPushToRepository.

jebrando (dismissed):
:shipit:

.github/workflows/two-human-reviewers.yml
```diff
@@ -0,0 +1,143 @@
...
+#
+# Gating on "the author is a bot" rather than on a list of named bots is deliberate: an app can be
+# renamed, and a stale name would silently let its pull requests through under-reviewed. Exempting a bot
+# is therefore an explicit act, and forgetting to update this file can only ever be too strict.
+#
+# It deliberately never checks out the PR's code, so running on `pull_request_review` (which carries a
+# writable token) cannot execute anything from the branch under review.
+
+on:
+  workflow_call:
```
darobs: Could an automated process remove this and void this check in its own PR?  I'm guessing having the default of at least one real human should help to prevent this.

mattdurak: I'm not sure, the bot can tell us.

But I think the goal here is not to protect from adversarial users, it's really just to "keep us/our team honest" when the bot submits PR's so someone who is not "driving" the task signs off.

As long as the existing checks guard against things like non-contributors etc., this is just additive.

If there are improvements we can make to make it more reliable/harder to bypass that's good too, but not a hard blocker. So Mr. Bot, look into this and fix if needed

.github/workflows/two-human-reviewers.yml
mattdurak: Fix if possible Mr. Bot

.github/workflows/two-human-reviewers.yml
mattdurak: Same as other comment, let's fix this if we can to tighten it up

darobs (approved):

